PR #8, opened 2026-05-13
fix: prevent consumer from overriding loop task status

youchengsong:
The previous commit's status restore logic in _consume_output fired on every subprocess exit, including dispatcher-managed loop iterations. This caused loop tasks to be marked "completed" after each iteration instead of continuing. Add a chat_initiated flag so only chat-initiated runs restore status in the consumer; dispatcher retains full control of loop/normal task lifecycle.


zjw49246:
Closing: duplicate of PR #7. Changes merged to main via commit 939b91d. Thanks!

PR #9, opened 2026-05-13
Fix/cancel kills process

youchengsong:
---                                                                           
  Summary                                                                       
                                                                                
  - Extract _stop_task_process() helper from stop-session endpoint and reuse it
  in cancel_task, so cancelling a loop/goal task also kills the running Claude
  Code process
  - Handle deleted tasks in dispatcher loop/goal checkpoints — previously only
  checked for cancelled status, now also stops when the task is removed from DB 
  
  Test plan                                                                     
                                                                  
  - Start a loop task, click cancel in the UI, verify the Claude Code process is
   killed
  - Start a goal task, click cancel, verify process stops
  - Delete a running loop/goal task, verify dispatcher stops iterating

### DIFF
--- a/backend/api/chat.py
+++ b/backend/api/chat.py
@@ -144,6 +144,7 @@ async def send_chat_message(
         git_env=git_env,
         thinking_budget=inst.thinking_budget,
         effort_level=effort_level,
+        chat_initiated=True,
     )
 
     task.status = "executing"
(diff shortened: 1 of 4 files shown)


zjw49246:
Closing: all changes (chat_initiated flag + cancel kills process + deleted task handling) merged to main via commit 939b91d with additional tests. Thanks!

PR #36, opened 2026-07-16
feat: update button fallback for non-systemd uvicorn

LianningLBY:
## Summary
- Update button now works for users who start CCM with `uvicorn` directly (without systemd)
- Detects whether the process is managed by systemd: if yes, uses `systemctl --user restart`; if not, falls back to kill + re-exec `uvicorn`
- Both restart paths (normal update + rollback) are covered

## Context
Many users run CCM via bare `uvicorn` without configuring a systemd service, causing the frontend update button to fail on the restart step.

## Test plan
- [x] systemd-managed service: restart via `systemctl --user restart` (unchanged behavior)
- [ ] bare `uvicorn` start: restart via kill + re-exec fallback

### DIFF
--- a/backend/services/update_service.py
+++ b/backend/services/update_service.py
@@ -7,7 +7,9 @@ import json
 import logging
 import os
 import shutil
+import signal
 import subprocess
+import sys
 import time
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


youchengsong (changes requested):
Good approach — detecting systemd vs bare uvicorn and branching is the right idea. A few issues to address:

### 1. Unused import (minor)
`import signal` is added (line 10 in the diff) but never used anywhere. Remove it.

### 2. `&&` chain after `kill` can leave the service dead (important)
In the fallback path:
```python
f"sleep 2 && kill {os.getpid()} && sleep 1 && "
f"cd {self.project_dir} && "
f"{sys.executable} -m uvicorn backend.main:app "
```
If the killed process exits with a non-zero code, `kill` itself returns 0 as long as the signal was delivered — so that part is fine. **But** if `kill` fails for any reason (PID already gone, permissions), the `&&` chain aborts and uvicorn is never restarted, leaving the service down with no indication.

Consider using `;` after the kill+sleep to make the restart unconditional:
```python
f"sleep 2 && kill {os.getpid()}; sleep 1; "
f"cd {self.project_dir} && "
```
This way the restart is attempted even if kill had an unexpected failure.

### 3. Hardcoded uvicorn args (minor, non-blocking)
The fallback restart drops any flags the user originally passed (e.g. `--reload`, `--workers`). This is acceptable for v1 but worth a comment noting the limitation so future readers understand it's intentional.

PR #39, opened 2026-07-16
feat(theme): 自定义主题颜色 + 背景图

LianningLBY:
## Summary
- 自定义主题颜色引擎：取色器选色 → oklch 色阶自动生成，覆盖全局 CSS 变量
- 背景图上传：IndexedDB 存储，强度滑块控制透明度，半透明表面透出壁纸
- 调深基础深色色阶（对标 Tailwind gray 默认暗度），降低 scrim 浓度和模糊，提升通透感

## Test plan
- [ ] 打开偏好菜单，选择「自定义」主题
- [ ] 用取色器选一个品牌色，确认全局色调变化
- [ ] 上传背景图，调节强度滑块，确认透出效果
- [ ] 切换回内置主题（深色/浅色），确认恢复正常
- [ ] 移动端测试布局无溢出


youchengsong (changes requested):
## Review findings

### 1. 🐛 SURFACE_ALPHA vs test expectations mismatch (customTheme.ts / customTheme.test.ts)

`SURFACE_ALPHA['900'] = 0.70` but the test asserts `0.66` at visible=100:

```ts
// customTheme.ts
'900': 0.70,  // 画布

// customTheme.test.ts
expect(alphaOf(vars['--color-gray-900'])).toBeCloseTo(0.66, 3);  // expects 0.66
```

`surfaceAlpha('900', 100)` = `1 - (100/100) * (1 - 0.70)` = **0.70**, not 0.66. Similarly, `'800': 0.88` but test expects 0.92, `'950': 0.85` but test expects 0.80. The midpoint test (`0.83 = 1 - 0.5*(1-0.66)`) also uses the 0.66 base. Either the constants or the test expectations need to be aligned — as-is the test will fail.

### 2. ⚠️ CSS `filter: none` contradicts blur/frosted-glass comments

The CSS comment block for `html[data-theme='custom'][data-has-bg]::before` says the pseudo-element should be **模糊**(blurred), and `customBg.ts` references "模糊柔和的背景" / "毛玻璃观感". But the actual rule sets `filter: none`. If the blur was intentionally removed, the comments should be updated; if it was accidentally dropped, the background image will show unblurred through the semi-transparent surfaces.

### 3. 💡 No-op test migration in alembic versions

`alembic/versions/e2e0716noop_test_update_flow_noop.py` looks like an E2E test artifact. It runs on every deployment via `init_db()` → `alembic upgrade head`. While it's harmless (empty up/down), it adds noise to the migration chain. Consider removing it or at least confirming it's intentional.

---

Backend update_service changes (systemd cgroup escape, rollback safety, `_is_managed_by_systemd` via `/proc/self/cgroup`) look solid with good test coverage. The shell script EXIT trap pattern is well-designed. Issue #1 above is the blocker.

youchengsong (approved):
LGTM - automated review passed. Clean fixes for systemd cgroup escape, SQLite rollback corruption, and orphan process detection. Custom theme engine is well-structured with thorough test coverage.

youchengsong (approved):
LGTM — thorough fix for the systemd cgroup escape + SQLite rollback corruption, and the custom theme engine is well-designed.

**Minor observations (non-blocking):**

1. **No-op migration file** (`e2e0716noop_test_update_flow_noop.py`): This E2E test artifact will be harmless in production but clutters the migration history — consider removing before merge if it was only for verification.

2. **Resource leak in `_spawn_update_script`**: The non-managed path does `stdout=open(log_file, "a")` without closing the file handle. Minor since the parent process is about to restart, but a `with` block or explicit close would be cleaner.

3. **CSS comment vs code**: The `::before` pseudo-element comment says "模糊它" (blur it) but `filter: none` applies no blur — the "frosted glass" effect comes from semi-transparent layers + scrim, not CSS blur. The comment could be clarified to avoid confusion.

4. **Base dark theme visual change**: gray-950 went from `oklch(16%)` to `oklch(7.5%)` and foreground from `98.5%` to `100%` — this darkens the default theme for all users, not just custom theme users. Presumably intentional ("对标 Tailwind gray 默认暗度"), just flagging it's a user-visible change.

Overall: well-tested (the trap test that reproduces the incident is excellent), good architectural decisions (IDB for images, oklch scale derivation from reference curves), clean separation between custom theme engine and existing theme system.

PR #42, opened 2026-07-17
fix: ask_user hook killed by CLI 600s default timeout — native AskUserQuestion picker froze PTY turn

Citrus086:
## Symptom

Task 32 (2026-07-17 13:36 UTC): the model called `AskUserQuestion`, the CCM card appeared — then vanished from the frontend a few minutes later. After that the turn never ended: zero model output, queued user messages never delivered, the claude process sat at **zero CPU with zero TCP connections** until `stop-session` manually tore it down.

## Root cause

The PreToolUse hook entry injected by `ensure_ask_user_hook` carries no `timeout` field, so the CLI kills the hook command at its **default 600s** — while the server side (`POST /api/ask-user/wait`) blocks up to `ask_user_timeout` = 1800s waiting for the user's answer:

1. Hook killed at 600s mid-wait → treated as no-decision → **native** `AskUserQuestion` proceeds.
2. In PTY mode the native tool renders an interactive picker waiting for keyboard input nobody can press → the turn freezes forever. ("Fail open to the native tool" is a safe fallback in `-p` mode, but a **deadlock** in PTY mode.)
3. The killed hook also disconnects the HTTP request → the server's `await` gets `CancelledError` → registry entry discarded → frontend pending backfill turns empty and the card silently disappears, so the user loses even the chance to answer late.

**Evidence for 600s (not 60s):** task 28's card on 07-14 was answered 3m42s after appearing and the answer fed back successfully — the hook survived well past 60s. The 600s default matches the CLI's Bash wall-clock cap.

## Fix (two parts)

- **`backend/services/ask_user_settings.py`** — the injected hook entry now carries an explicit `"timeout": ask_user_timeout + 60`, so the CLI never kills the hook before the server window closes.
- **`backend/hooks/ask_user_hook.py`** — inverted timeout semantics: a `timed_out` response now emits a PreToolUse **deny** with "No user response within the waiting window… proceed with your best judgment", instead of failing open. Only unreachable-CCM / unmanaged-session / unexpected errors still fail open to the native tool.

## Tests

`backend/tests/test_ask_user.py` (15 passed; full backend suite 1013 passed):
- injected entry carries the `timeout` field (= `ask_user_timeout + 60`)
- the real hook script is executed via subprocess against a stub `/wait` server for all three outcomes: `answered` → deny with the formatted answer; `timed_out` → deny with proceed-on-judgment; `no_session` → no output (fail open)

## Deployment note

Production is currently running a transitional hot patch (in-memory wrap of `ensure_ask_user_hook` + runtime `ask_user_timeout` lowered to 540s so pre-existing claude processes — whose hook snapshot still has the 600s default — get a server response before the kill; plus the same hook-script edit on disk). The in-memory patch dies at the next service restart and the disk edit will be reverted by the next rsync deploy, so **this PR should be merged before the next deploy** or the freeze regresses.

https://claude.ai/code/session_01WRLp8B5udzz1HUzz9ZJdGC


youchengsong (approved):
LGTM — well-diagnosed root cause (CLI 600s default hook timeout vs 1800s server wait), correct two-part fix (explicit timeout field + timeout→deny instead of fail-open), good test coverage including subprocess-level hook script testing. No correctness, security, or performance concerns.

youchengsong (approved):
LGTM - correctly fixes the 600s CLI timeout killing the ask_user hook. Tests are thorough.

youchengsong:
Code review passed ✅ 但与最新 main 有冲突（可能是刚合并的 #44 或 #48 造成的），请 rebase 解决冲突后我会合并。

```bash
git fetch origin main
git rebase origin/main
# 解决冲突后
git push --force-with-lease
```

youchengsong:
Merged to main locally after resolving conflicts (fork PR, rebased and merged manually).

PR #52, opened 2026-07-19
fix: support system service scope in self-update

zjw49246:
## Summary
- detect whether the running CCM service is managed by system or user systemd scope from its cgroup
- use the matching systemctl/systemd-run invocation for one-click update restart and migration scripts
- keep bare uvicorn fallback behavior unchanged
- harden update_migrate.sh tests by stubbing sudo so system-scope tests cannot escape the fake systemctl/systemd-run PATH

## Why
The cyf deployment runs as a system service under /system.slice, but the updater assumed user services and called systemctl --user. That allowed code/build updates to land while the live backend kept running the old commit, which hid newly added Codex GPT-5.6 models until a manual restart.

During validation, the full update_service test file initially interrupted the execution session because script integration tests could choose system scope in this CCM-hosted runner and call sudo -n systemctl outside the test stub. The PR now fixes that test isolation issue by stubbing sudo in the script test environment.

## Validation
- python -m py_compile backend/config.py backend/services/update_service.py backend/tests/test_service_update.py
- bash -n scripts/update_migrate.sh
- uv run pytest backend/tests/test_service_update.py -q

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -240,6 +240,11 @@ sudo systemctl stop ccm-backend      # 停止后端
 sudo journalctl -u ccm-backend -f    # 查看后端日志
 sudo journalctl -u ccm-tunnel -f     # 查看 tunnel 日志
 # 开机自启已通过 systemctl enable 配置
+
+# 自动更新的 systemd 服务配置（.env）:
+#   SERVICE_NAME=ccm-backend   # 服务单元名（不含 .service 后缀），默认 ccm
+#   SERVICE_SCOPE=auto         # auto | user | system，auto 从 cgroup 自动检测
+# 非默认服务名的部署必须显式配置，否则更新机制无法正确 restart
 ```
 
 ## 数据库
--- a/backend/api/system.py
+++ b/backend/api/system.py
@@ -21,7 +21,7 @@ _GIT_COMMIT: str = git_head_commit()
 
 @router.get("/health")
 async def health():
-    return {"status": "ok", "commit": _GIT_COMMIT, "test": "system-scope-update"}
+    return {"status": "ok", "commit": _GIT_COMMIT}
 
 
 @router.get("/stats")
--- a/backend/api/system.py
+++ b/backend/api/system.py
@@ -21,7 +21,7 @@ _GIT_COMMIT: str = git_head_commit()
 
 @router.get("/health")
 async def health():
-    return {"status": "ok", "commit": _GIT_COMMIT}
+    return {"status": "ok", "commit": _GIT_COMMIT, "scope_test": True}
 
 
 @router.get("/stats")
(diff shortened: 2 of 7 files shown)


youchengsong (approved):
LGTM - automated review passed. Clean refactor from boolean is-managed check to scope-aware detection. Auto-detection from cgroup text is sound, sudo -n usage is appropriate for system scope, and test coverage is good.

youchengsong (approved):
LGTM - automated review passed. Clean scope detection from cgroup, proper sudo -n usage for system services, and good test isolation with the sudo stub.

youchengsong (approved):
LGTM - automated review passed. Clean separation of scope detection logic, proper sudo -n for system services, and good test isolation with the sudo stub to prevent test escapes.

youchengsong (approved):
LGTM - automated review passed. Clean refactor from boolean _is_managed_by_systemd to scope-aware _systemd_scope. Good test isolation fix with the sudo stub.

youchengsong (changes requested):
One issue to fix before merging:

**`backend/api/system.py`: Debug artifact in health endpoint**

```python
return {"status": "ok", "commit": _GIT_COMMIT, "test": "system-scope-update"}
```

The `"test": "system-scope-update"` field is a development/debugging leftover and should be removed. The health endpoint response should remain:

```python
return {"status": "ok", "commit": _GIT_COMMIT}
```

Everything else looks good — the scope detection logic, sudo usage, shell script changes, test isolation, and config additions are all solid.

youchengsong (approved):
LGTM - automated review passed. Clean refactoring of systemd scope detection with proper test isolation (sudo stub prevents escape). The cgroup-based auto-detection logic is correct and the fallback chain is sensible.

youchengsong (changes requested):
One issue to fix before merging:

**`backend/api/system.py` — debug artifact in production health endpoint**

The diff adds `"scope_test": True` to the `/api/system/health` response. This looks like a debug/testing leftover and should be removed before merging. This is a public, unauthenticated endpoint — it shouldn't carry arbitrary debug flags.

```diff
-    return {"status": "ok", "commit": _GIT_COMMIT, "scope_test": True}
+    return {"status": "ok", "commit": _GIT_COMMIT}
```

Everything else LGTM — the scope detection, sudo -n usage, test isolation fix, and shell script helpers are all clean.

youchengsong (approved):
LGTM - automated review passed. Clean refactoring from bool to scope-aware detection, proper sudo -n usage for system services, good test isolation with sudo stub.

youchengsong (approved):
LGTM - automated review passed. Clean refactor from boolean  to scope-aware . Good test isolation with sudo stub, and the frontend sawDown guard correctly prevents false-positive reconnection to the pre-restart server.

youchengsong (approved):
LGTM - automated review passed. Clean refactor from boolean is_managed_by_systemd to scope-aware systemd_scope. Good test isolation with sudo stub, and the frontend sawDown guard correctly prevents false-positive reconnection to the pre-restart server.

PR #53, opened 2026-07-20
fix: support Codex same-turn message steering

Citrus086:
## Summary

- route the existing task injection endpoint by provider: Claude keeps PTY injection while Codex uses app-server `turn/steer`
- send `expectedTurnId` with every Codex steer so a turn-boundary race cannot attach input to a later turn
- expose the Codex app-server capability to the frontend and gate the injection UI by provider/transport
- reject unsupported Worker/shared-task injection explicitly and preserve a clear 409 for ended, non-steerable, or exec-fallback turns
- add protocol, API, runtime capability, and frontend regression coverage; update the agent documentation

## Why

The injection button was controlled only by the global Claude PTY setting. On Codex tasks it still called the PTY-only backend path, which searched the Claude PTY session pool using a Codex thread id and therefore could never inject successfully. With PTY disabled, the UI hid injection even though Codex app-server supports native same-turn steering.

## Verification

- real Codex 0.144.6 app-server E2E: started a turn running `sleep 8`, submitted `turn/steer` while the command was active, received `STEER_ACCEPTED=True`, and the same turn changed its final response to the injected marker (`STEER_OK_7F3A91`), exit code 0
- `uv run pytest -q backend/tests/test_service_instance_manager.py backend/tests/test_codex_app_server.py backend/tests/test_api_chat_plan.py backend/tests/test_api_settings_runtime.py` — 143 passed
- `npx vitest run src/components/Chat/ChatView.test.tsx -t "Live turn injection"` — 2 passed
- `npm run build` — passed

## Limitation

Codex `exec` fallback has no live steering transport. In that case the UI may still offer injection based on the configured app-server capability, but the endpoint returns 409 and retains the draft so the user can retry or send it as the next turn.


youchengsong (approved):
LGTM - automated review passed. Clean provider-based routing for injection, race-safe expectedTurnId on steer, proper gating for worker/shared tasks, and solid test coverage across all layers.

youchengsong (approved):

PR #56, opened 2026-07-22
fix: atomically reserve queued-message instances

LianningLBY:
## Summary

- Atomically select and reserve idle local instances across fresh dispatch and per-task queued-message consumers.
- Protect reservation release with token/owner validation.
- Requeue the exact QueuedMessage on InstanceAlreadyRunningError instead of dropping it.
- Document the incident, root cause, and regression coverage.

## Root cause

Different task queue consumers could select the same database-idle instance before either published its in-memory launching claim. The loser then hit InstanceAlreadyRunningError, while the generic consumer exception path acknowledged queue completion and silently lost that message.

## Tests

- Rebased dispatcher suite: 115 passed.
- Full backend suite before rebase: 1382 passed with repository-default environment overrides.
- Real frontend integration test on port 8010: sent simultaneous follow-ups from two task tabs; both ran on distinct instances, both rendered visible replies, the browser console had zero errors, and server logs contained no InstanceAlreadyRunningError.


youchengsong (approved):

youchengsong (changes requested):
发现一个 PostgreSQL 下的合并阻断：

`_reserve_idle_instance()` 把 `_running_tasks` 的全部 key 合并进 `busy_iids`，随后传给 `Instance.id.notin_(busy_iids)`。但远程 Worker 调度会在 `_running_tasks` 中写入 `worker-{task.id}` 这样的字符串 key；PostgreSQL/asyncpg 会把 `Instance.id` 的 bind 参数按 INTEGER 处理，字符串 key 会造成类型错误并打断本地 dispatch loop。SQLite 测试不会暴露这个问题。

请仅将整数 instance ID 纳入 SQL 过滤，并补充“存在活跃远程 Worker key 时仍可分配本地 instance”的回归测试。核心的原子预留及消息重排队方向是正确的；修正这个问题并解决当前 PROGRESS.md 的追加冲突后即可再审。

youchengsong (approved):
Reviewed correctness, concurrency, PostgreSQL compatibility, failure cleanup, and integration with maintenance admission. Fixed mixed string/integer running-task keys in the Instance.id predicate, added strict regression coverage, and released reservations when dispatch pauses after selection. 264 related backend tests pass; Alembic has one head.

youchengsong (approved):

PR #58, opened 2026-07-22
fix: exclude terminal instances from concurrency cap

Citrus086:
## Summary

- count only idle/running instances toward max_concurrent_instances
- apply the same capacity semantics during startup provisioning and runtime idle top-up
- preserve error/stopped rows for diagnostics and leave physical deletion to the existing cleanup endpoint
- add regression coverage for the production case where terminal rows already exceed the configured cap

## Root cause

Nine error/stopped instance rows remained in the database without owning processes. Both worker provisioning paths used the total number of database rows as the concurrency count, so max_concurrent_instances=8 prevented any idle worker from being created. New tasks and resumed chat messages then looped on No idle instance.

## Verification

- targeted dispatcher tests: 121 passed
- full backend suite: 1387 passed
- git diff --check


youchengsong (approved):

PR #59, opened 2026-07-22
fix: make self-update task-safe and notify users

luka7620:
## Summary

- Block self-update and rollback while task work, queued resumes, or prompt-only instance runs are active.
- Serialize every task-start path with the maintenance gate and make the final blocker check atomic with shutdown/spawn.
- Serialize update and rollback admission with one operation lock so concurrent requests cannot replace rollback state or start duplicate rollback scripts.
- Keep queued user/monitor resumes visible as blockers, while stop-session queue clearing removes obsolete bookkeeping without hiding in-flight work.
- Cancel the dequeue-to-in-flight handoff with a per-task queue generation, so a cleared message cannot launch after disappearing from the blocker model.
- Detect manual pulls by comparing the running commit with disk HEAD.
- Check for updates after page load and hourly, showing a dismissible non-blocking notice only when action is needed.

## Root cause

The original update flow coordinated only with the main Dispatcher dequeue path and queried only active Task rows. Per-task resumes, RalphLoop, manual launches, and prompt-only Instance runs could therefore escape the blocker model. Several restart paths still awaited broadcasts or grace periods after their final check, leaving a race before shutdown.

Queue cancellation originally left permanent pending markers. The first fix synchronized queue draining and in-flight bookkeeping, but a narrower handoff remained: `q.get()` removed a message before the consumer acquired the admission lock to increment `_task_queue_inflight`. A concurrent clear could therefore observe an empty queue and zero in-flight work, remove the blocker, and still allow the consumer-held message to launch.

Update start and rollback also used different admission synchronization. A rollback could pass its initial check, await dispatcher/task state, and then read a newly replaced update state with an empty or incorrect commit/backup. Two rollbacks could likewise pass the unlocked check together.

## Safety guarantees

- force never bypasses active-work protection.
- Task-query failures fail closed and never start the updater.
- Global dispatch, per-task user/monitor resumes, RalphLoop, and manual Task launches share one admission gate.
- Prompt-only runs persist a running Instance with no Task id and remain maintenance blockers until the instance is idle.
- Accepted queued resumes remain blockers before DB task status becomes active.
- stop-session queue clearing updates pending and in-flight bookkeeping under the same admission lock.
- Each queued message captures its task's cancellation generation. Clear advances that generation under the admission lock; a dequeued but unclaimed stale message is acknowledged and discarded without processing or launch.
- Work that already registered an in-flight claim remains a blocker until its current lifecycle finishes; clear cannot hide it.
- Update and rollback share one operation-admission lock. Rollback captures and validates its exact state, commit, and backup while holding that lock through the shutdown reservation decision.
- Concurrent update/rollback and rollback/rollback requests admit only one operation; a rollback script cannot be spawned with replaced state.
- Normal restart, migration restart, manual-pull fast restart, and rollback perform the final blocker query and synchronous shutdown/spawn in one admission-lock section, with no await after the successful final check.
- Dispatcher resumes whenever an update or rollback is rejected or cancelled.
- Dry-run caching never caches blockers; remote fetch failures cannot hide a local manual-pull restart requirement.
- Automatic checks are dry-run only and never pull code or restart the service.

## Verification

- Update/Dispatcher tests after rebasing current upstream main: 151 passed, 2 skipped, 5 deselected.
- Deterministic queue regressions cover enqueue -> clear before dequeue, clear after dequeue but before in-flight registration, and the registered in-flight control case.
- Deterministic operation regressions cover rollback paused while an update starts and two concurrent rollback requests.
- stop-session API tests: 9 passed.
- Chat/plan queue tests: 35 passed.
- Affected API, message pipeline, and RalphLoop tests from the prior review cycle: 199 passed, 1 deselected.
- UpdateButton component tests: 31 passed.
- Python compile checks for all modified backend and test modules.
- TypeScript no-emit validation passed.
- Production frontend build passed.
- git diff --check passed.

The five update-service deselections are Linux-only bash/systemd and /tmp script tests that cannot run correctly on this Windows host. The single API deselection is the existing task-ordering baseline assertion; neither is caused by this PR.

## Manual verification

- Verified the top update notice is dismissible and does not block normal page interaction.
- Verified manual update checks still open the detailed modal.
- Verified an executing task is listed as a blocker and update confirmation is disabled.
- Verified a forced update request still returns HTTP 409 while work is active.
- Verified the latest-version state remains silent during automatic checks.
- Verified the local frontend, backend health endpoint, and frontend API proxy return HTTP 200; a real update dry-run reported no phantom `queued_resume` blocker after the queue tests.

## Out of scope

The separate Codex rollout Windows UTF-8/GBK regression work remains intentionally excluded from this PR.


youchengsong (changes requested):
Blocking issue: the maintenance gate does not cover all paths that can start task work, and the final task check is not actually adjacent or atomic with shutdown.

GlobalDispatcher.pause_dispatching() only gates claims in _dispatch_loop. The per-task chat and monitor consumer still reaches the launch path and sets task.status = "executing" (currently around backend/services/dispatcher.py:3874-3878) without consulting _dispatch_paused or taking _dispatch_claim_lock. Therefore a message queued after the initial active-task check can start work during an update or rollback.

The new pipeline recheck around backend/services/update_service.py:748 does not close that race: both restart paths perform awaited broadcasts and a one-second sleep after the check before spawning or stopping, and the state.old_commit == state.new_commit && needs_restart fast path around lines 619-629 bypasses that recheck entirely. rollback() likewise checks only before its broadcast and sleep window. A queued chat or monitor turn can become executing in any of those windows and then be killed by the restart, contradicting the PR main safety guarantee.

Please make every task-start and launch path honor the same maintenance gate, including per-task queued resumes, or otherwise establish a shared atomic gate covering the last active-task check through shutdown. Add regression tests that pause the update at the pre-restart await and attempt a queued resume, plus coverage for rollback and the early fast-restart branch.

luka7620:
Addressed the requested task-start/shutdown races in commits 6be8dd0 and 4cd60aa. The shared admission gate now covers global dispatch, per-task user/monitor resumes, RalphLoop dequeue, and manual Task launches; accepted queued resumes remain visible as blockers. Fast restart (including the manual-pull branch), migration restart, and rollback now perform the final blocker query and synchronous shutdown/spawn in one admission-lock section, with no await after the final check. Added deterministic regressions for the preparation-to-launch race, queued user/monitor resumes at the pre-restart window, migration and non-migration paths, manual-pull fast restart, rollback, fail-closed checks, RalphLoop, and manual instance launch. Validation: 262 relevant backend tests passed (7 deselected), TypeScript no-emit validation passed, the production frontend build passed, and an isolated manual test confirmed an executing task still returns HTTP 409 even with force=true.

youchengsong (changes requested):
Two blocking correctness issues remain in the maintenance gate:

1. POST /api/instances/{instance_id}/run still permits a prompt-only launch (prompt set, task_id=None). The new task_start_guard() only protects that launch while it is being admitted; once instance_manager.launch() returns, no Task row has been marked in_progress/executing. UpdateService._get_active_tasks() only queries Task rows, so an update started while this prompt-only process is running sees zero blockers and may restart/kill it. Please either disallow prompt-only runs, represent them as task blockers, or include running taskless instances in the blocker query, and add coverage for starting an update after a prompt-only run has launched (not only for rejecting a task-id run while maintenance is already paused).

2. enqueue_message() adds the task id to _pending_task_starts, but clear_task_queue() can empty the queue without removing it. If stop-session clears a newly queued message before the consumer dequeues it, the consumer remains blocked in q.get(); after its 300-second idle timeout its finally block also does not discard the id. The stale id therefore remains forever and every later update/rollback reports a phantom queued_resume blocker. Please make queue clearing/cancellation update the pending-start bookkeeping under the same admission synchronization and add a regression test for enqueue -> clear before dequeue -> blocker query.

luka7620:
Addressed both remaining blocking issues after rebasing onto current upstream main.

- db1b219 adds running prompt-only Instances to the maintenance blocker query. The regression starts from an admitted prompt-only run whose Instance is persisted as running with no Task row, then verifies that start_update(force=True) is rejected and no pipeline is scheduled.
- db1b219 makes clear_task_queue asynchronous and synchronizes queue draining plus pending/in-flight bookkeeping with the admission lock. A per-task in-flight count prevents clearing queued messages from hiding work already dequeued. Consumer idle/cancellation cleanup also converges the blocker state. The regression executes enqueue -> clear before dequeue -> blocker query and verifies that no queued_resume blocker remains.
- c7b44ec updates README, CLAUDE, TEST, and PROGRESS with the final rebased commit ids and verification results.

Post-rebase verification: 147 update/Dispatcher tests passed (2 skipped, 5 Windows Linux-script deselections), 199 affected API/message/Ralph tests passed (1 existing ordering-test deselection), UpdateButton 31 passed, TypeScript no-emit validation passed, production build passed, Python compile checks passed, and git diff --check passed.

youchengsong (changes requested):
Blocking correctness issue: update start and rollback are not serialized by the same operation lock.

`start_update()` now holds `_start_lock` while it pauses dispatching, checks blockers, replaces `self._current` with a fresh state (whose `old_commit` is initially empty), and schedules the pipeline. `rollback()` does not take `_start_lock`; it checks the old `self._current.old_commit`, then awaits `_pause_dispatching()` and `_get_blocking_tasks()`, and only afterwards reads `self._current.old_commit` again. A concurrent `start_update()` can run during either await and replace `_current`. The rollback request can therefore proceed with the new state's empty/wrong commit and backup, race the new pipeline for `_lock`, and spawn the rollback shutdown script with incorrect arguments. Two concurrent rollback requests can similarly both pass the unlocked pre-check before either acquires `_lock`.

Please serialize both entry points with the same admission/operation lock, and capture/validate the rollback state while holding that lock. Keep the lock through the decision that reserves the operation (not necessarily through the full pipeline), so only one update or rollback can be admitted. Add a deterministic regression that pauses rollback after its initial check, starts an update, then resumes rollback and verifies that only one operation is admitted and no rollback script is spawned with the replaced state.

The maintenance-gate fixes requested in the earlier reviews otherwise appear covered. Targeted backend run: 235 passed; the two failures were unrelated environment overrides of DEFAULT_PROVIDER=claude.

luka7620:
Addressed the latest operation-admission review in `c2ab2dd`; `f6f41e0` records the invariant and regression coverage in the project docs.

- `start_update()` and `rollback()` now share `_operation_lock`.
- Rollback captures and validates the exact `UpdateState`, `old_commit`, and `backup_file` while holding that lock, and keeps admission reserved through the final shutdown/spawn decision.
- The rollback script validates state identity and only receives the captured commit/backup.
- `test_rollback_and_update_share_operation_admission_lock` deterministically pauses the first rollback after validation, starts an update, then verifies the update waits and is rejected after the rollback reservation; the script is spawned once with the original state.
- `test_concurrent_rollbacks_admit_only_one_operation` verifies two rollback requests cannot both pass admission or spawn duplicate scripts.

Verification after the change:

- Update/Dispatcher: 149 passed, 2 skipped, 5 Windows-only Linux-script deselections.
- System API: 14 passed.
- The two new deterministic concurrency regressions pass.
- Python compile, TypeScript no-emit, and `git diff --check` pass.

The earlier prompt-only Instance and queued-resume cleanup fixes remain unchanged. Please re-review the updated head `f6f41e0`.

youchengsong (changes requested):
Blocking correctness issue: `clear_task_queue()` can still hide a message that the consumer has dequeued but has not yet registered as in-flight.

In `_task_queue_consumer`, `q.get()` completes at backend/services/dispatcher.py:3635-3637, but `_task_queue_inflight` is not incremented until the consumer later acquires `_dispatch_claim_lock` at lines 3645-3649. In that lock-free window, `clear_task_queue()` can acquire the admission lock, observe `q.empty()` and an in-flight count of zero, and discard `_pending_task_starts` at lines 3588-3602. The consumer still holds the message and will subsequently register it and process it; therefore stop-session can report that the queue was cleared while the supposedly cleared resume later launches. If the updater acquires the final maintenance guard between the clear and the consumer registration, it can also see no blocker and commit shutdown while that accepted message is still held by the consumer.

The added `enqueue -> clear before dequeue` test does not exercise this ordering. Please make dequeue-to-in-flight ownership atomic from the perspective of queue clearing (or add a cancellation/generation mechanism so clear invalidates a dequeued-but-not-started item), and add a deterministic regression that pauses the consumer after `q.get()` but before in-flight registration, calls `clear_task_queue()`, and verifies that the message neither launches later nor disappears from the blocker model unless it was actually cancelled.

luka7620:
Addressed the dequeue-to-in-flight queue-clear race in `ab0d827`; `6245165` records the invariant and regression coverage in the project docs.

- Every `QueuedMessage` captures a per-task `queue_generation` while enqueue holds `_dispatch_claim_lock`.
- `clear_task_queue()` advances that generation under the same lock before draining the queue and reconciling pending/in-flight state.
- `_claim_dequeued_message()` validates the generation while holding the admission lock. A consumer-held message invalidated by clear calls `q.task_done()` and is discarded without entering `_process_queued_message()` or launching.
- A message that already registered as in-flight remains in `_pending_task_starts`; clear cannot hide that real blocker.
- A dequeued-but-unclaimed cancellation is counted as a cleared message, so stop-session reports success rather than returning a false “no running session” response.

The deterministic regression pauses the consumer after `q.get()` and before `_claim_dequeued_message()`, runs `clear_task_queue()`, resumes the consumer, and verifies that processing/launch never occurs and the blocker disappears only because the generation cancellation completed. A control regression verifies that already-registered in-flight work remains blocking until it finishes.

Verification:

- Update/Dispatcher: 151 passed, 2 skipped, 5 Windows-only Linux-script deselections.
- stop-session API: 9 passed.
- Chat/plan queue tests: 35 passed.
- Focused queue regressions: 3 passed.
- Python compile, TypeScript no-emit, and `git diff --check` pass.
- Local backend/frontend/proxy health checks return HTTP 200, and a real update dry-run reports zero phantom blockers.

Please re-review the updated head `6245165`.

youchengsong (changes requested):
Blocking correctness issue: active CCM monitor and one-shot sub-agent processes are still absent from the maintenance blocker model.

`UpdateService._get_blocking_tasks()` (backend/services/update_service.py:313-353) only combines active `Task` rows, taskless `Instance` rows, and queued resumes. However, `GlobalDispatcher._monitor_session_lifecycle()` and `_sub_agent_session_lifecycle()` launch independent subprocesses that can run for up to hours (`backend/services/dispatcher.py:3064-3190` and `3321+`). These processes are tracked in `_monitor_processes` / `_sub_agent_processes` and `SubAgentSession.status`, not as running `Instance` rows. A monitor is explicitly designed to outlive the parent turn, so the parent `Task` can already be `completed` while its monitor is still `running`.

In that state the update/rollback checks see zero blockers and commit the restart. A systemd restart kills the monitor/sub-agent with the service cgroup; on the next startup the stale running session is marked failed. In the bare-uvicorn restart path, `start_new_session=True` can instead leave the child orphaned while the old dispatcher that owns it disappears. Either outcome violates this PRs task-safe update guarantee.

Please include active CCM-owned monitor/sub-agent lifecycles in the blocker model (using DB status and/or the dispatcher process/task registries, with the final check under the same maintenance admission protocol). Add a regression where the parent Task is completed but a monitor or one-shot sub-agent remains running, and verify both update and rollback refuse to restart until that lifecycle finishes. Also cover the final-check path so such work cannot disappear between the initial check and shutdown commit.

youchengsong (approved):
LGTM after final review and maintainer fix: existing maintenance-gate races are covered, and cancellation during update/rollback admission now always reopens dispatch. Verified 243 affected backend tests, 31 UpdateButton tests, and TypeScript.

PR #57, opened 2026-07-22
fix: make PR review webhooks idempotent

LianningLBY:
## Summary

- Persist the PR head SHA and GitHub delivery ID on each PR review.
- Treat the same repository, PR number, and head SHA as one semantic review.
- Add database uniqueness constraints so concurrent webhook deliveries cannot create duplicate review tasks.
- Return the winning review after a concurrent uniqueness conflict instead of an HTTP 500.
- Keep reviewing a PR when its head SHA changes, while ignoring exact redeliveries.
- Add a reversible Alembic migration that preserves existing review rows.

## Root cause

The webhook handler only checked review status before creating a task. Synchronize events always created a new review, the payload head SHA and X-GitHub-Delivery header were not stored, and the check-then-insert flow had no database uniqueness guard.

## Tests

- PR Monitor API, service, and migration tests: 62 passed.
- Full backend suite: 1391 passed with test-only repository-default environment overrides.
- Frontend production build passed.
- Real test deployment on port 8010 migrated successfully. Replaying an already-processed delivery returned ignored, created zero new tasks, and all temporary verification rows were removed.

PR #60, opened 2026-07-23
feat: support Codex task distillation

LianningLBY:
## Summary

- route chat-page task Distill through the task provider while preserving the existing Claude path
- run Codex distillation as an isolated `codex exec --ephemeral` request using the bound or another healthy pool account without rebinding the task thread
- pass transcripts over stdin, parse Codex JSONL output, clean up timed-out processes, and expose provider/model metadata to the UI
- keep the latest `raw_content` transcript collection behavior from `main` and update the provider-aware UI copy/documentation

## Validation

- `pytest -q backend/tests/test_skill_distill.py backend/tests/test_api_chat_plan.py` — 41 passed
- `npx vitest run src/components/Chat/ChatView.test.tsx --reporter=dot` — 40 passed
- `npx tsc --noEmit --pretty false`
- Python compile check and `git diff --check`
- manual local service-level call with `gpt-5.6-sol` returned a complete five-section Skill card


youchengsong (changes requested):
Blocking issues:

1. `backend/services/skill_distill.py`: the new direct `codex exec` process bypasses `InstanceManager`'s per-CODEX_HOME admission/maintenance tracking. `_select_codex_distill_home()` only checks the pool's enabled/cooldown state; it cannot see `_codex_home_maintenance`, and the spawned process is not registered in `_codex_exec_homes`. Therefore a relogin/delete can consider the home idle and replace or clean `auth.json` while distillation is running, or distillation can start after maintenance has reserved the home. This breaks the repository's credential atomicity guarantees and can make a distill read partial/replaced credentials. Please route ephemeral distillation through an InstanceManager admission/tracking primitive (or add an equivalent per-home reservation that maintenance observes) and release it safely on every exit path. Add a concurrency test covering distill versus relogin/delete maintenance.

2. `distill_task_conversation()` does not clean up its subprocess on task/request cancellation. On Python 3.11+, `asyncio.CancelledError` is not caught by `except Exception`, so cancellation during `process.communicate()` leaves `claude`/`codex exec` running and unreaped until it exits on its own. The timeout path is covered, but cancellation/server shutdown is not. Catch cancellation explicitly, kill and reap with cancellation-shielded cleanup, then re-raise; add a cancellation test.

Validation run: `uv run pytest -q backend/tests/test_skill_distill.py backend/tests/test_api_chat_plan.py` => 41 passed.

youchengsong (approved):
LGTM after maintainer fixes: Codex distill now participates in per-CODEX_HOME maintenance admission, cancellation kills and reaps the subprocess before releasing the home, and untrusted transcript processing uses a read-only sandbox without user config/rules. Verified 177 backend tests, 40 ChatView tests, and TypeScript.

youchengsong (approved):
Revalidated after syncing merged PRs #57 and #59: 177 affected backend tests, 40 ChatView tests, TypeScript, single Alembic head, and diff check all pass.

youchengsong (changes requested):
Blocking lifecycle issue in backend/services/skill_distill.py: the timeout and generic-exception branches call _terminate_task_distill_process(process) directly, unlike the cancellation branch. If the request is cancelled while that helper is awaiting process.wait() (for example during shutdown immediately after the 5-minute timeout fires), the wait is cancelled, codex_home_exec_guard exits, and CODEX_HOME maintenance admission is released before the child has been confirmed dead and reaped. I reproduced this: kill() was called, process.wait() remained blocked/cancelled, and the guard finalized. This violates the new invariant documented in CLAUDE.md and can overlap credential maintenance with a still-live process as well as leave a zombie. Please use the cancellation-shielded kill/reap path for every cleanup branch (especially TimeoutError and the generic exception path), and add a regression test that cancels during timeout cleanup and asserts the guard remains held until wait() completes.

youchengsong (changes requested):
Blocking findings:

1. **Security — `read-only` is not a confidentiality boundary** (`backend/services/skill_distill.py:116-133, 269-285`). The transcript and `custom_instruction` are untrusted user-controlled text, but `codex exec` still has shell/tool access. `--sandbox read-only` prevents writes; it does not prevent reading the host filesystem. A prompt injection can therefore ask Codex to read server-side files (for example credentials under the service user home) and place their contents in the generated card returned by the API. The natural-language instruction “不调用工具、不读取文件” cannot enforce this boundary, especially because the custom instruction and transcript follow it in the same user prompt. Please run distillation through a tool-free inference path, or place the process in an OS/container isolation boundary whose readable filesystem contains no host secrets (only the required input/runtime). Add an adversarial integration test demonstrating that a transcript/custom instruction cannot read a sentinel file outside the isolated input area.

2. **Correctness/security — timeout cleanup only kills the CLI parent** (`backend/services/skill_distill.py:183-206, 269-300`). The process is not started in its own session/process group, and `_terminate_task_distill_process` calls only `process.kill()`. If Codex has spawned a shell command (including through the injection above), killing/reaping the CLI can orphan that child, while the CODEX_HOME guard is then released. Please launch the ephemeral execution in an isolated process group/container and terminate + reap the whole group/workload on timeout and cancellation; cover this with a real child-process cleanup test.

Validation performed: `pytest -q backend/tests/test_skill_distill.py backend/tests/test_api_chat_plan.py` (43 passed), `vitest` ChatView suite (40 passed), and `tsc --noEmit` passed.

PR #65, opened 2026-07-24
feat(codex-pool): support email-only OTP login

Citrus086:
## Summary

- allow Codex account add and relogin with only an OpenAI email address
- switch the existing browser login to OpenAI's email/one-time-code path when available
- keep the same login process and OAuth/PKCE state while the administrator enters the six-digit OTP in CCM
- show a clear password-required failure when OpenAI does not offer an email-code option
- update the account modal so mailbox token and OpenAI password are both optional

## Safety

- OTP values remain in memory/stdin only and are never persisted or logged
- the existing login journal, `auth.json` validation, fsync barrier, maintenance lock, and rollback flow are unchanged
- email-only accounts persist only empty credential metadata; relogin requests a fresh human OTP

## Verification

- `uv run pytest backend/tests/test_codex_login_mailbox.py -q` — 92 passed
- `npx vitest run src/components/Layout/PoolDrawer.test.tsx` — 32 passed
- `npm run build` — passed
- production smoke: email-only login committed successfully, live quota read succeeded, and an ephemeral Codex turn on the new account returned the expected response


youchengsong (approved):

youchengsong (approved):

PR #61, opened 2026-07-23
refactor(mcp): introduce provider-neutral MCP server specs

luka7620:
## Summary

- introduce a provider-neutral `McpServerSpec` for CCM MCP servers
- build main-task, monitor, and sub-agent configs from shared server specs
- preserve the existing Claude `mcpServers` JSON and launch behavior
- pin each role's enabled-tool set to its actual FastMCP registrations

## Why

CCM currently builds Claude-specific MCP JSON directly in three separate paths.
Adding Codex app-server and exec support on top of those paths would duplicate
task/session context, authentication, paths, tool scopes, and timeout policy.

This PR separates "what server should run" from "how a provider receives that
server", giving later Codex PRs one shared source of truth.

## Scope

Included:

- immutable `McpServerSpec`
- provider-neutral builders for `ccm_skills`, `ccm_monitor_agent`, and
  `ccm_sub_agent`
- Claude JSON renderer
- compatibility and role-isolation tests

Not included:

- Codex app-server injection
- `codex exec` config rendering
- runtime/UI/capability changes
- third-party MCP management

## Compatibility

The existing `generate_mcp_config()`,
`generate_monitor_agent_mcp_config()`, and
`generate_sub_agent_mcp_config()` entry points are unchanged.

The Claude renderer emits the same `command`, `args`, `cwd`, and optional `env`
shape as before. Provider-only metadata is intentionally not added to Claude
JSON.

## Tests

- 47 MCP config/server/monitor tests passed
- 5 Claude command-builder tests passed
- main/monitor/sub-agent spec snapshots
- enabled tools checked against actual FastMCP registrations
- exact Claude JSON compatibility for all three roles
- auth, API base, Windows/Linux paths, duplicate names, and immutability
- manual verification: 13 main tools, 3 monitor tools, 3 sub-agent tools

## Follow-up

The next PR will render these specs for Codex app-server and `codex exec`
without changing task launch behavior yet.

PR #66, opened 2026-07-24
feat(mcp): render CCM server specs for Codex

luka7620:
## Summary

- render provider-neutral `McpServerSpec` values for Codex app-server thread config
- render the same specs as `codex exec -c` argv without shell quoting or global config writes
- preserve command, args, cwd, env, required, enabled tools, and timeout policy
- validate Codex server-name and timeout constraints before process launch

## Why

PR #61 introduced a shared MCP server spec while intentionally leaving Codex disconnected. This follow-up adds the pure Codex adapters needed by later app-server and exec integration PRs, without changing task launch behavior yet.

During CLI smoke testing, a per-server override such as `mcp_servers.name.with.dot=...` was found to be split by Codex's dotted-path parser even though it round-tripped through Python `tomllib`. The exec renderer therefore emits one merged `mcp_servers={...}` inline table, which both tested Codex versions deep-merge with existing user MCP entries.

## Scope

Included:

- app-server `config.mcp_servers` renderer
- `codex exec -c` TOML argv renderer
- Windows paths with spaces, Unicode, quotes, backslashes, env maps, arrays, and multiple servers
- duplicate/invalid name and timeout validation
- no-write guarantees for `$CODEX_HOME/config.toml`
- docs and verification notes

Not included:

- injecting MCP config into production Codex turns
- changing app-server, exec fallback, retry, or account-rotation behavior
- UI or capability changes

## Verification

- `65 passed` — `test_mcp_config.py` + `test_mcp_server.py`
- `5 passed` — Claude command-builder compatibility tests
- `14 passed` Monitor regression tests; 2 Windows-only local runs were blocked by the existing hard-coded `/tmp` log path
- `npx tsc --noEmit` passed
- `compileall` and `git diff --check` passed
- Codex CLI `0.144.6` and `0.145.0`: isolated-home `mcp list -c ... --json` smoke tests passed
- both app-server versions successfully completed `thread/start` with a real CCM FastMCP server configured as required
- no smoke test created or modified `config.toml`

## Follow-up

A later PR will inject these specs into `thread/start`/`thread/resume`; another will use the exec renderer in fallback and own MCP process cleanup.


youchengsong (approved):

youchengsong (changes requested):
P2 correctness: backend/services/mcp_config.py:269-271 assumes every JSON basic-string escape is valid TOML, but json.dumps(..., ensure_ascii=False) leaves U+007F (DEL) literal and TOML rejects that character. A McpServerSpec containing DEL in command, args, cwd, env keys/values, or enabled_tools therefore produces an invalid mcp_servers=-c override; Codex may then treat the failed TOML value as a raw string and fail MCP initialization. Reproduced with tomllib for command, args, and env while all 56 existing test_mcp_config.py tests pass. Please use a TOML-correct string/key encoder (including escaping U+007F as  without corrupting non-BMP Unicode) and add a regression test.

PR #68, opened 2026-07-24
fix: compact and continue Codex tasks after context overflow

LianningLBY:
## Summary

- preserve Codex app-server `contextWindowExceeded` errors and model context-window metadata instead of flattening them into generic system text
- normalize current-context token accounting for app-server and exec fallback, using the latest rollout request rather than cumulative turn totals
- compact task history and automatically continue both queued chat and fresh/mode lifecycle paths in a new Codex session with the original message
- keep the existing proactive threshold behavior accurate for Codex's effective context window

## Why

Codex tasks can reach the 272K context limit and exit even though CCM is expected to summarize the existing conversation and continue. The structured app-server error was not recognized by every lifecycle path, and cumulative token accounting could also produce false utilization values above 100%.

## Testing

- `11 passed` — newly added context-overflow regression tests
- `141 passed, 275 deselected` — Codex/context-focused dispatcher, instance-manager, app-server, and compaction tests after rebasing onto latest `main`
- `1981 passed` — isolated backend full suite before the final rebase
- isolated frontend integration on port 8011:
  - seeded usage at `210,000 / 258,400` (81%, threshold 80%)
  - UI displayed the automatic compaction/new-session notice
  - the new thread returned `E2E_CONTEXT_CONTINUED`
  - task finished `completed`, with no error and context reset to about `1.2K / 272K`

The existing port 8010 test environment was not modified during the isolated integration run.


youchengsong (approved):

PR #69, opened 2026-07-25
fix: only invoke Skills for explicit commands

LianningLBY:
## Summary

- treat `enabled_skills` as availability metadata instead of injecting fake `$command` invocations into every initial prompt and follow-up
- append invocation instructions only when the user message actually starts with a registered `$command`
- keep quoted or mid-message references such as “你怎么知道 `$monitor`？” as ordinary text
- preserve temporary per-turn Skill enablement for real `$monitor` / `$sub-agent` commands

## Verification

- `uv run pytest -q backend/tests/test_api_chat_plan.py backend/tests/test_service_dispatcher.py` — 224 passed
- test deployment smoke task #95 (`provider=codex`, Monitor/Sub-Agent enabled): initial prompt, ordinary follow-up, and mid-message `$monitor` mention all returned the requested sentinel only, with no fake Skill invocation

The branch is rebased on the latest `origin/main` and contains one commit touching only the prompt construction, regression tests, and the project convention note.


youchengsong (approved):

PR #72, opened 2026-07-26
feat: support Codex Sub-Agents with safe thread cleanup

LianningLBY:
## Summary

- add Codex support for the existing CCM Sub-Agent workflow using independent app-server threads
- inject required task-scoped MCP tools for parent and child threads
- preserve account binding and shared app-server isolation
- interrupt and delete one-shot child threads without stopping the shared transport
- unsubscribe resumable parent threads after terminal turns
- support explicit initial task commands without treating enabled skills as invoked

## Safety and lifecycle

- child thread deletion retains cleanup evidence on failure
- manual stop interrupts only the matching turn
- child rollout and MCP helper are removed after completion or stop
- parent thread history remains resumable and unloads after the Codex idle grace period

## Validation

- backend related suite: 615 passed
- frontend component tests: 28 passed
- live test Task 103: normal completion, result injection, child helper and rollout removed
- live test Task 104: manual stop, child helper and rollout removed

Tested on the isolated CCM test service; production was not changed.


youchengsong (changes requested):
There is one blocking lifecycle issue:

- **Parent Codex threads launched by the normal fresh-task lifecycle are never unsubscribed.** `_launch_codex_app_server()` only records `enabled_skills` in `_launch_params` when `chat_initiated` is true (lines 1331-1344), while the new terminal cleanup at lines 3640-3645 decides whether to call `thread/unsubscribe` exclusively from `_launch_params`. Fresh dispatcher launches use the default `chat_initiated=False`, including initial `$delegate` launches, so their lookup is empty and the unsubscribe is skipped. This leaves the task-scoped `ccm_skills` stdio MCP helper/subscription loaded after every initial terminal turn, contrary to the lifecycle guarantee in this PR and potentially leaks one helper per task. Please carry the controller-MCP fact on the turn/process or otherwise make it available for all launches, and add a regression test that exercises a real non-chat launch rather than manually seeding `_launch_params`.

Validation: the affected backend suites pass (515 tests), but the current unsubscribe test manually populates `_launch_params`, which masks the fresh-launch path above.

youchengsong (approved):
LGTM after lifecycle fix. Fresh and chat Codex turns now carry exact thread-scoped MCP cleanup ownership on the turn handle, with regression coverage for both MCP and non-MCP paths. Validation: 516 affected backend tests passed, 28 frontend tests passed, frontend production build passed, and 6 focused post-fix tests passed.

youchengsong (approved):

PR #73, opened 2026-07-27
feat: add native Codex session forks

LianningLBY:
## Summary

- add native Codex `thread/read`, `thread/fork`, and empty-thread creation through the account-scoped app-server registry
- expose one Fork action beside Inject for completed Codex sessions
- list only persisted user messages as valid fork anchors; injected, tool, model, and system events are excluded
- fork before the selected user message, then open a new Task with that message prefilled but not automatically sent
- preserve the source account/home and working directory while keeping the new Codex thread independently resumable

## Safety and lifecycle

- serialize read/fork/create/delete operations with active turns, rebinds, and account maintenance
- settle cancellation after mutating fork RPCs so the created thread is never orphaned silently
- compensate failed database persistence with native `thread/delete`
- retain the newly merged Codex Sub-Agent delete/unsubscribe lifecycle while integrating Fork

## Validation

- rebased onto current `origin/main`; merge-tree reports no conflicts
- backend Fork/app-server suite: 108 passed
- ChatView component suite: 44 passed
- verified in the isolated CCM test environment


youchengsong (changes requested):
Requesting changes:

- **Correctness — preserve attachments when replaying the selected fork message.** `list_codex_fork_anchors` exposes the selected message's attachments, but `fork_codex_task` persists only `fork_seed_message`, and `ChatView` hydrates only the text input. `handleSend` derives `file_paths` exclusively from the current `useFileUpload` state, so neither an initial prompt's attachments nor a follow-up user message's attachments are sent from the forked task. Because the native thread is deliberately forked *before* the selected user message, those files/images are absent from native history too; sending the prefilled draft silently changes the model input. The initial-anchor path also leaves copied attachment metadata on a task with `description=None`, where it is not rendered or submitted. Please carry the seed attachment/path data into the fork composer (or replay it safely server-side on the one-shot seed send) and add backend/component coverage for both initial and follow-up anchors with attachments.

Validation performed: backend targeted suite 108 passed; ChatView suite 44 passed; frontend production build passed.

youchengsong (approved):
LGTM after maintainer fixes: fork seed attachments are preserved in the editable composer, and all native thread operations use the unified CODEX_HOME app-server admission guard. Validation: 373 backend tests, 46 ChatView tests, frontend production build, compileall, and diff check passed.

PR #74, opened 2026-07-27
feat(codex): preserve required MCP in exec fallback

luka7620:
## Summary
- wire the existing Codex exec MCP renderer into the production main-task launch path
- retry required main-task MCP through an equivalent `codex exec` only when app-server fails before `turn/start`
- keep unsafe replay fail-closed, keep Sub-Agent app-server-only, and serialize app-server/exec ownership per canonical `CODEX_HOME`
- document the rollout flag, safety boundary, and manual verification flow

## Validation
- 53 focused Codex MCP/transport tests passed
- Python compileall and `git diff --check` passed
- frontend `npx tsc --noEmit` passed
- real `codex-cli 0.145.0` / `gpt-5.6-sol` smoke through production `_build_command`: `ccm_skills/ccm_command_help` completed and the turn returned `PR4_EXEC_MCP_OK` with exit code 0

## Existing Windows-only baseline failures
The broader related files still contain pre-existing Windows incompatibilities around `signal.SIGKILL`, POSIX `0700` mode assertions, `/tmp` path normalization, and the Linux-only `fcntl` deployment guard. The PR4-focused matrix is green and these failures do not touch the changed launch paths.


youchengsong (changes requested):
There is one correctness issue in the new per-CODEX_HOME transport exclusion.

`InstanceManager.read_codex_rate_limits()` still calls `registry.read_rate_limits()` directly, without taking `_codex_home_lock()` or checking `_codex_exec_homes`. A forced/manual pool refresh reaches this callback (`CodexPool.fetch_quota(..., live=True)`) and `CodexAppServerRegistry.read_rate_limits()` creates/starts an app-server when none exists. Therefore, after an exec generation has spawned and registered its home, a quota refresh can start an app-server against the same `CODEX_HOME` concurrently. The reverse launch checks and the idle shutdown added here do not cover this path.

Please route live quota reads through the same home admission/ownership gate (reject or otherwise safely serialize while an exec generation owns the home), and add a regression test that starts/records an exec generation and verifies a concurrent live quota read cannot start an app-server. This is important because the PR explicitly relies on one transport owner per canonical home to avoid shared auth/session state races.

luka7620:
Addressed the requested CODEX_HOME ownership fix in 1d154cf. Live quota reads now normalize the home, acquire the same per-home admission lock, reject recorded task/ephemeral exec ownership before creating the registry, and hold the lock through the quota RPC. Ephemeral exec admission also shuts down an idle app-server or rejects an active one. Added four regressions covering recorded exec rejection, ephemeral exec rejection, RPC/exec serialization, and active app-server rejection. Validation: Windows focused 4 passed; Linux ownership matrix 10 passed; adjacent app-server/pool/distill suites 135 passed; real-child-process smoke confirmed quota:start -> quota:end -> app-server:shutdown -> exec:enter. Documentation update: 4491ae9.

youchengsong (changes requested):
There is one remaining correctness issue in the new per-`CODEX_HOME` transport exclusion.

`GlobalDispatcher._launch_codex_sub_agent()` still calls `_ensure_codex_app_server_registry().start_turn()` directly (`backend/services/dispatcher.py:7158-7169`). This path does not take `InstanceManager._codex_home_lock()` and does not call `_assert_codex_app_server_home_available()`. Consequently, after a normal exec generation has registered the home in `_codex_exec_homes`, or while task distillation holds `_codex_ephemeral_home_users`, a queued Codex sub-agent can create/start an app-server against that same home. It can also race in the gap after an exec shuts down the idle server and registers ownership, because the registry itself has no knowledge of the manager-level exec maps. This violates the PR’s one-transport-per-home invariant and can race shared auth/session state.

Please expose a manager-level app-server admission helper that canonicalizes the home, holds the same home lock, checks maintenance/normal-exec/ephemeral-exec ownership, and performs `registry.start_turn()` atomically; route both main-task and Codex sub-agent app-server launches through it. Add regression coverage proving `_launch_codex_sub_agent` cannot start an app-server while (1) `_codex_exec_homes` owns the selected home and (2) `codex_home_exec_guard()` has an active ephemeral user.

The focused Codex MCP/transport test selection otherwise passed locally (96 passed).

youchengsong (approved):
LGTM after maintainer fixes: unified CODEX_HOME app-server admission now covers main tasks, live quota reads, and Codex sub-agents; 318 targeted tests passed.

youchengsong (approved):

PR #75, opened 2026-07-27
fix: hide context copy for Codex tasks

LianningLBY:
## Summary

- show **Copy context from** only for Claude tasks
- hide the unsupported copy control when Codex is selected
- add a regression test covering the provider switch

## Testing

- `npx vitest run src/components/Tasks/TaskForm.test.tsx` (26 passed)
- `npm run build`

### DIFF
--- a/frontend/src/components/Tasks/TaskForm.test.tsx
+++ b/frontend/src/components/Tasks/TaskForm.test.tsx
@@ -307,6 +307,22 @@ describe('TaskForm copy-context-from select overflow fix', () => {
     expect(screen.queryByText('Copy context from:')).not.toBeInTheDocument();
   });
 
+  it('does not show any copy-from control for Codex', async () => {
+    vi.mocked(api.listTasks).mockResolvedValue([
+      { id: 10, description: 'task', session_id: 'sess-1', title: null, project_id: 1 },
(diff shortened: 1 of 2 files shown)


youchengsong (changes requested):
Blocking correctness issue in `frontend/src/components/Tasks/TaskForm.tsx`:

The new provider guard only hides the copy-context control. It does not clear or gate `cloneFromTaskId`. A user can select a source while the form is set to Claude, switch the provider to Codex, and submit; `handleSubmit` still includes `clone_from_task_id`. The backend then clones the Claude session JSONL and assigns that cloned Claude session ID to the new Codex task, leaving the task with a provider/session mismatch.

Please either clear `cloneFromTaskId` when switching away from Claude or, more defensively, only include `clone_from_task_id` when `provider === "claude"`. Add a regression test that selects a source, switches to Codex, submits, and verifies `api.createTask` receives no `clone_from_task_id`; the current test only verifies that the control is hidden.

youchengsong (approved):
LGTM after fixing the provider-gated submission path and adding regression coverage. TaskForm tests and frontend build pass.

youchengsong (approved):

PR #77, opened 2026-07-28
feat(codex): enable task-scoped Skills and User Skills

luka7620:
## Summary

- build one provider-aware, task-scoped Skill context for Claude and Codex, including selected User Skill discovery with on-demand full-content reads
- preserve Skill selections and authoritative User Skill snapshots across app-server, safe exec fallback, retry/resume/fork, Worker forwarding, and migration
- enforce Codex capability gates in the UI and API: Monitor remains unsupported, the main-MCP kill switch allows only Sub-Agent, command-required Skills are rejected before persistence/queueing, and Worker snapshots never fall back to colliding local IDs

## Validation

- frontend PR6 tests: `31 passed`; `npm run build` passed
- backend API/MCP/Codex group: `301 passed`; the 7 remaining failures reproduce existing Windows path/signal/task-sort differences
- `test_skill_context.py`: `6 passed`
- new InstanceManager paths: `5 passed`
- Worker Relay: `72 passed`
- authoritative-snapshot regression matrix: `10 passed`; adjacent MCP/Worker/migration matrix: `16 passed`
- Python compileall and `git diff --check` passed

Manual integration coverage:

- unselected User Skill is denied
- Fork inherits selected User Skills
- `codex exec` fallback preserves the same Skill/MCP context
- main MCP kill switch rejects User Skills while retaining Sub-Agent
- `$monitor` is rejected on create, description update, and follow-up before any task/message persistence

## Environment notes

A monolithic backend run was bounded at 15 minutes on Windows because lifecycle tests did not terminate, so the affected suites were rerun in file and targeted groups. The full frontend run was `431 passed, 3 failed`; the three failures reproduce existing Windows timezone/path-separator assumptions outside this change.


youchengsong (changes requested):
Two blocking issues:

1. `backend/api/tasks.py` performs `task_migrator.migrate(...)` before validating the effective provider/Skill configuration. A combined PUT such as changing `worker_id` together with `provider=codex` and `enabled_skills.monitor=true` migrates (and remotely recreates) the task first, then `_validate_skill_configuration` raises 400. That leaves externally visible state changed despite a rejected request and violates the stated pre-persistence gate. Please compute and validate the effective Skill configuration before entering the migration branch, and add a regression test asserting the migrator is not called for an invalid combined update.

2. With `CODEX_MAIN_MCP_ENABLED=false`, `InstanceManager` correctly exposes only the Sub-Agent controller spec, but that spec still enables `ccm_read_skill`. The updated `ccm_read_skill` checks only provider support, not the task selection / main-MCP capability, so Codex can call `ccm_read_skill("code-review")` (or any other non-Monitor ordinary Skill) while the API/UI claim only Sub-Agent is available. Please gate ordinary Skill reads to the effective selected/allowed Skills (while preserving the `sub-agent` read needed by the controller), and cover the kill-switch path with a test.

youchengsong (changes requested):
Blocking correctness issue in backend/api/tasks.py update_task:

The new pre-migration validation fixes rejection-side effects, but a valid combined update is still non-atomic when the destination is a remote Worker. update_task calls task_migrator.migrate(task_id, target) before queue.update_task applies provider/enabled_skills/selected_user_skills. TaskMigrator._ensure_worker_task therefore serializes the old Task configuration to the destination Worker. After migration returns, queue.update_task changes only the Manager mirror. For example, moving a Claude task to Worker 42 while setting provider=codex and enabling a Codex skill leaves the Worker copy as Claude with the old skill selection, while the Manager reports Codex/new skills. The follow-up skill sync does not repair provider, and other execution paths can observe the stale remote copy.

Please make the migration and effective configuration update one coordinated operation so the destination import receives the validated final provider/Skill tuple, with rollback/CAS behavior if the migration fails. Add a regression test for a valid combined local-to-Worker (and ideally Worker-to-Worker) update that asserts the destination payload and persisted Manager task contain the same final provider, enabled_skills, selected_user_skills, and snapshots. The existing test only covers the invalid-request/migrator-not-called case and misses this split-brain path.

I ran the affected backend groups at the PR head: 536 passed. The PR is also currently reported by GitHub as conflicting with main and will need to be rebased/resolved before it can merge.

youchengsong (changes requested):
Two blocking correctness issues remain at the current head:

1. `backend/api/tasks.py:update_task` still performs `task_migrator.migrate(...)` before applying non-routing Skill updates. A valid PUT containing `worker_id` plus `enabled_skills` and/or `selected_user_skills` does not enter `_update_*_task_with_routing_config` (those helpers are keyed only by provider/model/tier), so the migrator serializes the old Skill tuple and snapshots to the destination. Only after migration returns does `queue.update_task` update the Manager mirror. The destination Worker can therefore execute with stale Skills while the Manager reports the new selection. Please coordinate the final validated Skill configuration with migration (or reject all such mixed updates before any side effect), and add local-to-Worker and Worker-to-Worker regression coverage asserting both mirrors and the destination payload agree.

2. The new authoritative snapshot abstraction is bypassed in both `TaskMigrator._ensure_worker_task` and `WorkerProxy._user_skill_snapshots`: both call `build_user_skill_snapshot_payload(..., task.selected_user_skills)` without `metadata=task.metadata_`. For a task that was itself imported with `ccm_user_skill_snapshots`, an onward migration/forward queries the node-local `UserSkill` table instead of using the authoritative snapshot. A missing local row makes the next import fail; a colliding local ID silently substitutes the wrong skill body. Pass the task metadata through in both paths and cover the collision/missing-row cases.

I ran the affected backend review group at `8ecb21f`: 643 passed; these paths are not covered by the current tests.

youchengsong (changes requested):
Blocking correctness issue: remote Worker Skill-only updates are still split-brain outside the chat path.

In `backend/api/tasks.py:update_task`, `_WORKER_ROUTING_CONFIG_FIELDS` contains only provider/model/tier, so an update to `enabled_skills` and/or `selected_user_skills` on an already-forwarded task falls through to `queue.update_task` (around lines 1882-1915) and changes only the Manager mirror. The new `WorkerProxy.sync_task_skill_selection` is called only from `_send_worker_chat`; other execution admissions do not use it. In particular, `POST /api/tasks/{id}/retry` (around lines 2948-2962) checks routing and immediately proxies the retry to the Worker, whose task row still has the old Skill selection/snapshots. A user can therefore toggle a Skill on a completed Worker task and press Retry: the Manager/UI reports the new selection, but the Worker executes the retried turn with the old context. This also contradicts the PR claim that selections are preserved across retry/resume.

Please coordinate Skill-only Worker updates with the executable Worker row, or synchronize and prove the authoritative Skill tuple/snapshots at every remote execution admission (including retry and any analogous resume/plan paths). Add a regression test that updates `selected_user_skills`/`enabled_skills` on an already-forwarded task, retries it, and asserts the Worker receives the new selection and authoritative snapshots before execution.

luka7620:
已按三轮评审意见全部修复，请重新评审：

- `cce3721`：有效 provider/Skill 配置在任何 Worker 迁移/路由副作用前完成校验；非法组合更新不调用 migrator。
- `cce3721`：`ccm_read_skill` 限定为 Task 已启用/always Skill；主 MCP kill switch 关闭时只保留已选 `sub-agent`。
- `99d625c`：`worker_id + provider/Skills/User Skill snapshots` 改为协调迁移。目标 Worker 的 inert import 使用已校验的最终配置；导入成功后，Manager 才在同一次 CAS 中提交 Worker 指针与最终配置。失败保持 Manager 原配置，认领/完成 CAS 同时比较待更新字段原值。
- `99d625c`：`TaskMigrator._ensure_worker_task` 与 `WorkerProxy._user_skill_snapshots` 均传入 `metadata=task.metadata_`，确保 onward migration/forward 使用 authoritative Manager snapshot。
- `ff5093d`：补齐 local→Worker、Worker→Worker、失败回滚、并发 CAS，以及本地 User Skill 缺失/同 ID 内容碰撞测试；destination payload 与 Manager 的 provider、enabled_skills、selected_user_skills、snapshots 均逐项比对。
- `8ecb21f`：已合并最新 `upstream/main` 并解决冲突，PR 不再落后旧基线。

验证：相关后端矩阵 `292 passed, 9 deselected`（9 项均为已记录的 Windows 路径/SQLite 排序差异）；核心评审聚焦测试、前端 45 项测试和生产构建均通过；`compileall` / `git diff --check` 通过。

youchengsong (changes requested):
Blocking correctness issue remains at ff5093d: Skill-only edits on an already-forwarded Worker task are not synchronized to the executable Worker row.

`update_task` only routes fields in `_WORKER_ROUTING_CONFIG_FIELDS` (`provider`, `model`, `codex_service_tier`) through the Worker coordination path. An update containing only `enabled_skills` and/or `selected_user_skills` therefore falls through to `queue.update_task`, changing only the Manager mirror. `sync_task_skill_selection` is called before Worker chat, but not before other execution admissions. In particular, `retry_task` and `approve_plan` call `_ensure_worker_routing_ready` and then immediately proxy `/retry` or `/plan/approve`; routing readiness does not cover the Skill tuple or authoritative User Skill snapshots. The Worker can consequently execute the retried/approved turn with its old Skills while the Manager/UI reports the new selection.

Please make remote Skill updates coordinated with the Worker executable row (including authoritative `ccm_user_skill_snapshots`) under the task operation/generation fence, or synchronize and verify that tuple before every remote execution admission. Add regression coverage for an already-forwarded task that updates ordinary/User Skills and then retries (and plan-approve if synchronization is admission-based), asserting the Worker receives the final enabled_skills, selected_user_skills, and authoritative snapshots before its status becomes executable.

Validation performed at this head: the targeted backend group passed (307 tests), but no test exercises this Skill-only update followed by a non-chat Worker admission.

luka7620:
Addressed the latest execution-admission review in efc5924. Worker Skill-only saves now share the per-task operation lock with execution admission. Retry and Plan Approve now synchronize Manager-authoritative enabled_skills, selected_user_skills, and User Skill snapshots, require an exact Worker readback confirmation, re-check the Worker generation, and only then proxy the executable transition. Chat uses the same strengthened synchronization confirmation. Added targeted coverage for successful confirmation, stale confirmation fail-closed behavior, post-update Retry, post-update Plan Approve, and save/admission lock serialization. Also ran three adjacent routing-marker/retry-ABA regressions. Result: 8 targeted tests passed; no full suite was run.

youchengsong (changes requested):
Blocking correctness issue in the default Codex app-server path:

`backend/services/codex_app_server.py` lines 1066-1072 puts the canonical Skill directory in `turn/start.additionalContext`, but pinned Codex CLI 0.144.6 does not define that field on `v2/TurnStartParams`. I generated the protocol schema with `codex app-server generate-json-schema`; `TurnStartParams` contains `threadId`, `input`, cwd/model/effort/tier/etc., but no `additionalContext`. More importantly, the app-server silently ignores unknown request fields: sending the same invalid-thread `turn/start` request with no extra field, with this `additionalContext`, and with an arbitrary `totallyUnknownField` produced the identical thread-id validation response. It does not reject the field, so the new rejection/fallback logic never runs.

Consequently, the normal app-server route starts successfully but drops the ordinary/User Skill catalog. Only `codex exec` fallback receives it via `wrap_skill_context`, meaning the advertised task-scoped Codex Skills are unavailable on the primary transport. The new test only asserts that the mocked `_request` received the invented key, so it cannot detect protocol deserialization dropping it.

Please carry the context through a supported 0.144.6 protocol surface (for example, prefix the text element in `input` with the same bounded canonical context used by exec, or use another schema-backed mechanism), and add an integration/protocol-level regression that proves the model-visible turn input actually contains the catalog rather than merely checking the outbound Python dict.

youchengsong (changes requested):
Two blocking correctness issues remain at the current head:

1. `WorkerProxy.sync_task_skill_selection` requires the Worker response `status` (and `instance_id`) to exactly equal the Manager mirror before retry/chat/plan approval. That invariant is false for tasks moved by `TaskMigrator`: `/migration-import` deliberately creates the destination row as inert `cancelled`, while `_finish_migration` restores the Manager to its pre-migration status (for example `completed`, `failed`, `plan_review`, or `pending`). Therefore the first retry/chat/plan approval after a normal completed-task migration updates the remote Skills successfully, then fails the confirmation with 409 because remote `cancelled != completed`; the task can no longer resume. The new sync tests synthesize `status: task.status`, so they hide this real migration state. Please reconcile/prove the two intended generations without requiring these deliberately divergent statuses to match (or coordinate the destination status safely), and add an end-to-end regression for migrating a completed/plan-review task and then retrying/chatting/approving it.

2. `frontend/src/components/Tasks/TaskBadges.tsx` treats any `getRuntimeSettings()` failure as `codex_main_mcp_enabled=false` and caches that result for the page lifetime. The toggle handler then rebuilds `enabled_skills` by retaining only entries present in the filtered `tools` list. On a transient settings request failure, a Codex task with `code-review` enabled shows only Sub-Agent; toggling Sub-Agent silently sends a payload that deletes `code-review` (and every other ordinary Skill). Capability uncertainty must not cause an unrelated toggle to destructively clear persisted selections—preserve unknown/current keys or disable the control until capability is known. Please add a failure-path test.

Validation: the focused Skill/MCP/Codex-app-server/migration/Worker group passes at this head (280 passed), but neither scenario above is covered.

youchengsong (changes requested):
Blocking correctness issue: the new leading-$command capability gate is bypassed for Worker-backed chat (and likewise for shared-chat proxying).

`send_chat_message` dispatches to `_send_worker_chat` before it parses the command or calls `_validate_skill_configuration`. `_send_worker_chat` then commits and broadcasts the Manager-side `LogEntry` before proxying `/chat` to the Worker. For a Codex Worker task, sending `$monitor ...` therefore persists a visible user message on the Manager first; only afterward does the Worker reject the unsupported command with 400. This violates the PR requirement that incompatible command-required Skills be rejected before Task/LogEntry persistence and queueing, and leaves a ghost message for a turn that never ran. The same early branch means `_send_shared_chat` persists locally before the owner can reject the command.

Please parse/validate the leading command against the authoritative task provider before any local Worker/shared persistence or broadcast (while retaining the Worker-side validation as defense in depth). Add regression coverage for a Codex Worker task using `$monitor` that asserts no Manager `LogEntry` is written, no broadcast/proxy occurs, and the response is 400; cover the shared path too if shared tasks can carry Codex provider state.

youchengsong (changes requested):
Blocking correctness issue: the initial Manager→Worker dispatch can still execute with a stale Skill tuple.

`_update_worker_task_with_skill_configuration` only serializes the Manager-side save with `get_task_operation_lock`, but `Dispatcher._dispatch_worker_tasks` claims a pending Worker task as `in_progress` before taking that lock and passes the already-loaded `claimed_task` object into `WorkerProxy.forward_task_to_worker`. The proxy takes the lock later but never reloads the Task inside it. This leaves two bad interleavings:

1. After the Dispatcher claim but before the proxy acquires the lock, a Skill update can acquire the lock and commit the new `enabled_skills` / `selected_user_skills`; the later forward still serializes the stale pre-update `claimed_task` and the Worker executes the old context.
2. If forwarding acquires the lock first, the destination task can be created and start executing with the old tuple; the Skill update then acquires the lock and changes only the Manager mirror while the active Worker turn keeps the old context.

The retry/chat/plan preflight synchronization does not cover this initial dequeue/forward admission. Please either include the exact claim plus a fresh Task reload and destination creation in the same operation fence, or reject/coordinate Skill edits once this generation has been claimed so the destination payload is proven to match the final Manager tuple before it becomes executable. Add a deterministic regression test that pauses initial forwarding around the operation lock, updates ordinary/User Skills in each ordering, and asserts the Worker import cannot execute with the stale selection/snapshots.

youchengsong (approved):

PR #79, opened 2026-07-29
refactor(monitor): schedule provider-neutral monitor turns

luka7620:
## Summary

- replace the long-lived Claude Monitor loop with durable, one-check scheduled turns
- persist provider, next-run time, generation ownership, and bounded failure state
- fence every callback to the exact active generation and require one successful callback per turn
- recover safe sleeping schedules after restart while failing uncertain active generations closed
- coordinate stop, task cancellation/deletion, MCP config cleanup, and terminal state transitions
- keep the runner seam provider-neutral for PR7B while retaining the existing Claude-only product capability in PR7A

## Correctness details

- due turns are claimed with a database CAS that advances `turn_generation` and publishes `active_turn_generation`
- generation-specific MCP configuration carries the same token to `report_status` / `mark_complete`
- duplicate, stale, or tokenless callbacks are rejected without consuming a newer turn
- process exit without a callback is a failed turn; failures retry with bounded exponential backoff and terminate after three consecutive failures
- restart recovery only rehydrates local CCM Monitor rows with no active generation and an existing parent task
- startup stale-child cleanup now defers local CCM Monitor rows to Dispatcher recovery, including the normal case where the foreground parent task already completed

## Validation

- 191 focused Monitor/API/MCP/model/dispatcher tests passed
- 64 focused task termination/cancellation tests passed
- 3 focused startup recovery/fail-closed regression tests passed after the manual restart finding
- frontend targeted tests and TypeScript checking passed
- Alembic upgrade/downgrade/upgrade cycle passed
- Python compile checks and `git diff --check` passed

Manual Windows validation with a real Claude API account covered:

- consecutive ordinary and important reports followed by `mark_complete`
- no duplicate turn after terminal completion
- sleeping Monitor recovery across a backend process restart
- manual stop with no later generation
- exact `max_checks` completion
- 5s then 10s failure backoff and terminal failure on the third attempt
- shortened local timeout harness proving exact child-process reap

The full repository test suite was intentionally not run; validation was limited to the affected groups.

## Scope

This is PR7A only. Codex Monitor thread/app-server support and enabling Monitor for Codex remain in PR7B.


youchengsong (approved):
LGTM - full repository impact review passed; 365 focused tests and Alembic head validation passed

PR #80, opened 2026-07-29
feat(codex): add persistent monitor runtime ownership

luka7620:
## Dependency

- Stacked on #79 (`refactor(monitor): schedule provider-neutral monitor turns`).
- Please merge/review #79 first; this PR's diff will reduce to PR7B1 after that merge.

## Summary

- run internal Codex Monitors on dedicated durable app-server threads while preserving the PR7A scheduled-turn state machine
- persist exact thread/home/account identity and frozen model/effort/tier/cwd routing for restart recovery and cleanup
- inject only the required generation-fenced `ccm_monitor_agent` callbacks under a read-only isolated profile
- interrupt/delete the exact Monitor thread on Stop/terminal transitions without killing the shared account app-server
- keep public Codex Monitor creation disabled; capability/API/UI opening remains PR7B2

## Correctness details

- commit new thread identity before publishing its turn adapter or allowing `turn/start`
- retain durable cleanup evidence when compensation or terminal `thread/delete` fails, and retry it after restart
- block account retirement and parent Task deletion while a persisted/in-flight Monitor owner exists
- preserve a running schedule across safe CCM/app-server restarts, but fail closed on incomplete or uncertain identity
- after every reaped nonterminal check, settle `thread/archive` then `thread/unarchive` so the same thread/history rebuilds the next generation's callback MCP client
- terminalize and clean the exact thread if runtime recycling fails instead of allowing a stale generation callback
- disable project config, writes, autonomous agent features, Apps, and inherited MCP Apps for the Monitor thread

## Validation

- 289 focused Monitor/API/model/dispatcher/app-server/MCP tests passed in the development container
- 2 POSIX orphan-process cases passed in the same image with `--init`, so PID 1 could reap their deliberate child processes
- 11 focused account/task ownership and Alembic upgrade/downgrade/schema tests passed
- 6 targeted `MonitorPanel` tests passed
- TypeScript and Vite production build passed
- real Codex app-server check preserved one thread while refreshing callback MCP generation 101 to 102
- manual Windows CCM validation covered two checks on the same thread across backend restart, read-only rejection, `mark_complete`, and active-turn Stop
- active Stop cancelled generation 4, deleted the exact Monitor thread, cleared durable cleanup state, and left the shared app-server running

The full repository test suite was intentionally not run; validation was limited to affected groups.

## Scope

This is PR7B1: internal Codex Monitor runtime ownership and lifecycle only. Public Codex Monitor creation/API/UI capability remains explicitly disabled for PR7B2. Worker Monitor also remains closed.


youchengsong (approved):
LGTM — reviewed persistent Codex monitor ownership, generation fences, cleanup/account blockers, migrations, and UI integration. Focused backend suite, Monitor UI tests, compile checks, and production build passed.

PR #81, opened 2026-07-29
feat: add independent Plan task workflow

zjw49246:
## Summary

- model every Plan as an independent mode=plan Task, optionally related to a target Task through plan_target_task_id
- add a strictly read-only Planner/Reviewer pipeline with structured output, audit runs/steps, transient retry, account routing, CloudRouter admission, and exact process cleanup
- make approve/reject control-plane-only operations: neither wakes nor mutates the target session
- let users explicitly attach approved related Plans to the next real chat message; standalone Plans create an idempotent execution Task
- add staleness checks for conversation/repository changes, Worker routing and migration fences, multi-Plan limits, revision history, and Manager-local audit IDs
- add ChatView Plan history/composer attachments and update the global Plan review panel
- update the design, README, TEST, PROGRESS, and synchronized CLAUDE.md/AGENTS.md guidance

## Design decision

Approval does not append directly to a Claude/Codex native session. CCM persists the approved Plan as a pending attachment, and the model sees it only when the user explicitly selects it with a real message. This avoids an ACK turn, surprise execution, and active-session races while keeping in-session and New Task Plan flows consistent.

## Verification

- backend affected matrix: 490 passed
- frontend Chat/Plan tests: 77 passed
- TypeScript and Vite production build passed
- Python compileall, Alembic head/upgrade-downgrade checks, and git diff check passed

The full backend run has 13 failures. The exact same 13 tests fail in a clean origin/main@4249605 worktree, covering stale Monitor assumptions, auth/deployment environment assumptions, and update-script baselines; they are not introduced by this PR. Repository-wide ESLint also retains 59 pre-existing errors, with no new error in the changed files.

## Manual acceptance still required

Run the documented development smoke flow with real Claude and Codex accounts, then repeat the key path on a Worker before production deployment. This PR does not restart services or apply the migration to production.


youchengsong (approved):
LGTM after fixes — merged current main, repaired Alembic ancestry, restored Plan application on pre-admission shutdown, added exact Dispatcher/Ralph Plan cancellation and retained-process update blockers, closed generic Plan-limit bypasses, and made Codex Plan Standard-only to prevent silent Fast downgrade. Focused backend matrices (737 + 65 + follow-up regressions), 131 frontend tests, SQLite migration upgrade/downgrade/upgrade, compile checks, and production build passed. Full backend: 3038 passed before final fixes; the only non-baseline failure was corrected and targeted tests pass.

PR #82, opened 2026-07-29
feat(codex): enable local Monitor capability

luka7620:
## Dependency

- Stacked on #80 (`feat(codex): add persistent monitor runtime ownership`).
- #80 is itself stacked on #79 (`refactor(monitor): schedule provider-neutral monitor turns`).
- Please review/merge #79 and #80 first; this PR's diff will then reduce to PR7B2 only.

## Summary

- enable Monitor for confirmed local, non-Shared Codex Tasks when the main CCM MCP capability is enabled
- centralize the exact Task-scope decision across Task create/update, chat commands, Monitor APIs, MCP discovery/spec rendering, and Instance launch
- keep Worker, Shared, migration-managed, and snapshot-managed Codex Tasks fail-closed before persistence, proxy, or process side effects
- expose the derived deployment capability through Runtime Settings and update Task form, badges, chat, and Monitor panel behavior
- preserve existing and unknown Skill keys when Runtime Settings are temporarily unavailable

## Correctness details

- use one `codex_monitor_supported_for_scope` decision for backend capability checks instead of removing the previous blanket rejection independently at each entry point
- treat Worker management markers and User Skill snapshots as authoritative remote-scope evidence even when `worker_id` is temporarily absent
- recheck capability after the Monitor route write barrier so a concurrent local-to-Worker migration cannot create a local Monitor row
- rebuild the exact main MCP spec from the current Task generation at launch rather than trusting stale caller state
- advertise and read the Monitor Skill only for the same confirmed local scope used by the public API
- fail closed in the UI while capability is unknown without destructively clearing persisted selections

## Validation

- 118 focused backend Task/Chat/Monitor API, Runtime Settings, Skill context, MCP server/spec, Instance launch, and PR7B1 lifecycle tests passed in the Linux development container
- 127 focused frontend tests passed across Skill capabilities, Task form/badges, Chat view, and Monitor panel
- TypeScript/Vite production build passed
- `git diff --check` passed
- real local Codex manual validation completed five ordinary/important Monitor reports and reached `completed`
- active Monitor Stop cleared scheduling, deleted the exact Codex thread, left no cleanup error, and kept the shared account app-server alive

The full repository test suite was intentionally not run; validation was limited to affected groups.

## Scope

This is PR7B2: local Codex Monitor capability and UI completion. Worker and Shared Codex Monitor remain explicitly unsupported.


youchengsong (approved):
LGTM — reviewed exact local/non-shared Codex Monitor capability gates across Task/Chat/MCP/API/UI. Fixed the stale cancellation mock to cover terminal cleanup semantics. 615 focused backend tests, 127 frontend tests, and production build passed.

PR #83, opened 2026-07-30
Fix stale task busy detection after instance reuse

youchengsong:
## What changed

- reject stale in-memory consumer ownership when an instance has been durably reassigned to another task
- add a regression test covering a completed task whose old instance slot is reused by a new live task

## Root cause

The queued-chat busy check could combine a stale consumer record for the completed task with the live process of the new task occupying the same instance. That made the completed task appear permanently busy and caused every chat message to be requeued after 120 seconds.

## Impact

Completed tasks can resume chat normally after their former worker instance is reused. This fixes the production Task 211 failure mode without weakening the existing foreground and detached-generation fences.

## Validation

- `pytest -q backend/tests/test_service_dispatcher.py` — 212 passed
- `git diff --check`

### DIFF
--- a/backend/services/dispatcher.py
+++ b/backend/services/dispatcher.py
@@ -10637,6 +10637,20 @@ Codex 中工具会显示为上述 mcp__ccm_monitor_agent__* canonical 名称；
             return False
         instance_id = task.instance_id
 
+        instance = await db.get(Instance, instance_id, populate_existing=True)
+        if (
+            instance is not None
+            and instance.current_task_id is not None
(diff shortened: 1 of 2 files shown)


youchengsong:
Reviewed the current diff: the durable `Instance.current_task_id` ownership check correctly prevents stale consumer bookkeeping from being combined with a reused instance generation, and the regression covers the reported failure mode. I did not merge because this PR is still marked Draft; switch it to Ready when the remaining author-side work is complete.

PR #84, opened 2026-07-30
feat: add task-scoped artifact downloads

luka7620:
## Summary

- turn relative, absolute, and `file:` links in Task and Loop chat Markdown into one-click task artifact downloads
- enforce Task ACLs, canonical workspace containment, regular-file checks, a 100 MB limit, `last_cwd` resolution, and `/workspace` mapping
- stream Worker-hosted artifacts through the Manager while preserving download headers
- keep web URLs and other external links on their existing browser behavior

## Validation

- `.venv/bin/pytest -q backend/tests/test_api_task_artifacts.py` — 7 passed
- `TMPDIR=/private/tmp .venv/bin/pytest -q backend/tests/test_api_files.py backend/tests/test_worker_relay_proxy.py backend/tests/test_api_resource_acl.py` — 154 passed
- `npx vitest run src/api/client.task-artifact.test.ts src/components/Chat/ChatView.test.tsx src/components/Chat/LoopChatView.test.tsx` — 97 passed
- `npm run build`
- live end-to-end check: a Codex task generated a Unicode Markdown artifact; the endpoint returned `200 OK` with an encoded `Content-Disposition`, and the downloaded SHA-256 matched the source


youchengsong (changes requested):
Blocking security issue in `backend/api/task_artifacts.py`: the containment/type/size checks are TOCTOU checks on a pathname, but `FileResponse` opens that pathname later. Task processes are explicitly able to modify their workspace, so after `candidate.resolve()`, `is_file()`, and `stat()` succeed, a process can replace the checked file (or an ancestor directory) with a symlink before Starlette opens it. The download can then read a service-accessible file outside the task workspace. The same reopen window means a file can grow after the 100 MB check and the response can stream more than the advertised limit.

Please open the artifact through an anchored, no-follow path traversal (or an equivalent safe primitive), validate the already-open regular-file descriptor with `fstat`, and stream from that exact descriptor with an enforced byte cap; do not hand the checked pathname to a response class that reopens it. Keep the descriptor lifetime tied to response completion/cancellation. Add deterministic regressions that swap the final file and an intermediate directory after resolution but before streaming, plus a growth-after-check case, and assert that no outside bytes or bytes beyond the limit are returned. The Worker route inherits the destination Worker endpoint behavior, so fixing the endpoint closes both paths.

youchengsong (changes requested):
One containment race remains in the hardened implementation: `_open_workspace_root` calls `os.open(root, O_NOFOLLOW|O_DIRECTORY)` on the complete absolute pathname. On POSIX, `O_NOFOLLOW` protects only the final pathname component; symlinks in parent components are still followed. Because task processes run as the service user with write-capable/danger-full-access workspaces, a writable ancestor can be swapped before this open so `root_fd` points at an attacker-chosen outside directory, then restored before `_resolve_artifact_parts` performs `Path.resolve()` / `relative_to(root)`. The lexical check then succeeds against the restored workspace while `_open_resolved_artifact` traverses the already-open outside root FD and streams outside bytes.

Please anchor the workspace root itself with component-by-component no-follow traversal from a stable trusted directory descriptor (or an equivalent kernel primitive such as `openat2` with `RESOLVE_BENEATH`/no-symlink constraints), and derive/validate artifact traversal against that same anchored object rather than mixing an unanchored root FD with later pathname resolution. Add a deterministic regression that swaps a parent component of the configured workspace root around root-FD opening and restoration, and proves an identically named outside artifact is never returned. The final-file/intermediate-directory FD traversal and bounded streaming changes otherwise address the previous review.

youchengsong (approved):
LGTM - security races fixed, conflicts resolved, and targeted backend/frontend validation passed

PR #88, opened 2026-07-31
fix: preserve reverted Alembic revision history

luka7620:
Production DBs already at b6e1f4a2c9d7 cannot update because the published migration was deleted. This restores that exact revision, adds f7a1c3d9e5b2 as a forward cleanup migration for the reverted Plan schema, adds upgrade/downgrade regression coverage, and documents immutable published revisions. Verification: backend/tests/test_alembic_migrations.py 15 passed; one Alembic head; restored file SHA-1 matches history; git diff --check passed. Production validation: transactional update from cf6f53c to 3509cca completed, database advanced from b6e1f4a2c9d7 to f7a1c3d9e5b2, health reports the new commit, worktree is clean, and a mode-600 SQLite backup was retained.


youchengsong (changes requested):
Blocking migration issue against current `main`:

This branch restores the published `c8f5d3a72b10 -> b6e1f4a2c9d7 -> f7a1c3d9e5b2` lineage, while current main already contains the separately published `c8f5d3a72b10 -> 5f7a9c2e4d61` lineage. Resolving the Git conflicts alone would leave two Alembic heads, and neither published revision may be rewritten.

Please rebase/merge current main, preserve the published b6/f7 migration contents, and add a new no-op merge revision whose `down_revision` is `("f7a1c3d9e5b2", "5f7a9c2e4d61")`. Add upgrade coverage from clean DB, b6, f7, and 5f to the single new head (plus downgrade/re-upgrade where supported).

The PR is currently `CONFLICTING`, so it is not safe to merge until that graph and the conflicts are resolved.

youchengsong (approved):
复审通过：保留已发布 b6/f7 migration 原始字节，新增 7e4b9c1d2a63 no-op merge revision 汇合 main 的 5f sibling head；覆盖 clean、b6、f7、5f 升级及 merge/Plan cleanup downgrade-reupgrade。完整 backend/tests/test_alembic_migrations.py：22 passed；alembic heads 唯一；diff check 与 py_compile 通过。

PR #89, opened 2026-08-01
feat: render Markdown math with KaTeX

zxiangx:
## What changed

- add a shared `MarkdownRenderer` backed by `remark-math`, `rehype-katex`, and KaTeX
- render math consistently in task chat, Loop chat, shared chat, and Discussion views
- support native `$...$` / `$$...$$` delimiters and Codex-style `\(...\)` / `\[...\]` delimiters
- preserve formula-like text inside inline code and fenced code blocks
- keep unmatched delimiters as ordinary readable Markdown instead of consuming the rest of a message
- add responsive overflow styling for long display equations

## Why

Codex commonly emits display equations such as:

```text
\[
\nabla_z L_{\text{SFT}}=p_s-e_y
\]
```

CCM previously passed this through `react-markdown` with only `remark-gfm`. Markdown treated the backslashes before brackets as escapes, so users saw ordinary brackets and raw LaTeX commands instead of rendered equations.

The delimiter normalization happens before Markdown parsing and only converts complete pairs outside code. KaTeX then renders the resulting math nodes without enabling trusted HTML commands.

## User impact

Scientific and technical answers now display inline and block equations correctly across all chat surfaces. Existing Markdown behavior, task artifact links, code-copy controls, and code samples remain unchanged.

## Validation

- `113 passed` across the new renderer tests and Chat/Loop/Shared chat regression suites
- `npx tsc --noEmit`
- `npm run build`
- production smoke verified KaTeX CSS/font assets and the public health endpoint


youchengsong (changes requested):
Blocking findings:

1. `normalizeMathDelimiters` rewrites raw Markdown before parsing and only protects code spans/fences. It therefore mutates delimiter-looking bytes inside link destinations and inline HTML. Reproduction: `[docs](https://example.com/\(section\))` becomes `[docs](https://example.com/$section$)`, and `<span data-value="\(literal\)">` is mutated too. This violates the stated “existing Markdown behavior remains unchanged” contract and can break real URLs. Please tokenize/parse contexts so only Markdown text nodes are normalized, and add regressions for link/image destinations, autolinks, and HTML.

2. PR #90 is now merged and adds `remarkTaskArtifactPaths` to task chat. Rebase onto current `main` and make the shared renderer accept/preserve the task-only plugin; otherwise the artifact-link behavior will be lost or conflict in `ChatView`/`LoopChatView`.

The existing focused suite (113 tests) and build pass, but an isolated two-case regression for item 1 fails deterministically.

youchengsong (commented):
Additional blocking regressions found in adversarial rendering tests:

- `remark-math` keeps `singleDollarTextMath: true`, so ordinary currency such as `The starter plan costs $5 and the pro plan costs $20 per month.` renders the words between the two prices as math. A general-purpose chat renderer needs an explicit currency-safe policy.
- KaTeX is configured without `maxSize`; `$\rule{1000000em}{1000000em}$` emits DOM styles with million-em dimensions and can make the chat page unusable. Set a bounded `maxSize` and keep `trust: false`.
- The dependency tree contains direct `katex@0.18.1` while `rehype-katex` resolves `katex@0.16.47`; align the runtime/CSS version to avoid duplicate weight and compatibility drift.
- The raw preprocessor also mutates image destinations, autolinks, link-reference definitions, indented code, and can pair delimiters across paragraphs. This reinforces that the fix should be token/AST-aware rather than more global regular-expression exclusions.

youchengsong (changes requested):
Rechecked the current rebased head. The task-artifact plugin integration from #90 is now preserved, so that earlier integration concern is resolved. The rendering blockers remain unchanged:

- `normalizeMathDelimiters` still rewrites the raw Markdown string while protecting only backtick code. Delimiter-looking text in link/image destinations, autolinks, reference definitions, indented code, and inline HTML is still mutated before parsing; delimiters can also pair across Markdown block boundaries.
- `remark-math` still uses single-dollar inline parsing, so ordinary currency with two prices can turn the intervening prose into math.
- `rehype-katex` still has no bounded `maxSize`, allowing adversarial formulas to emit enormous dimensions.
- The lockfile still installs direct `katex@0.18.1` while `rehype-katex` resolves nested `katex@0.16.47`, so CSS/runtime versions remain split.

Please keep normalization AST/context-aware, define a currency-safe dollar policy, bound KaTeX output, align the KaTeX dependency, and add adversarial regressions before marking the PR ready.

youchengsong (approved):
复审通过：raw Markdown rewrite 已替换为 AST/context-aware 解析；URL/image/autolink/reference/HTML/code 与跨段落均隔离，单美元公式关闭以保护货币文本，KaTeX maxSize=20/trust=false，依赖统一为唯一 katex@0.16.47；保留 task-artifact plugin 与最新 main。Markdown 专项 10 passed、前端全量 523 passed、production build 与改动文件 ESLint、diff check 全通过。

PR #91, opened 2026-08-03
fix(codex): keep native Goals attached across turns

zxiangx:
## Summary

- keep one CCM `CodexTurnProcess` attached while an active native Goal continues across turns
- preserve continuation output, Task execution state, WebSocket delivery, and deterministic stop behavior
- safely adopt an active Goal left detached by an older CCM version through race-safe `turn/steer` reconciliation
- keep Fast-tier and non-Goal active threads fail-closed

## Problem

CCM previously detached at the first `turn/completed`, while Codex Goals immediately launched more native turns. The UI turned green, continuation output was dropped, and queued follow-ups retried until they happened to hit a turn-boundary window.

## Tests

- `pytest backend/tests/test_codex_app_server.py -q` (162 passed)
- `pytest backend/tests/test_service_instance_manager.py -q` (284 passed)
- `python -m compileall` for changed Python files
- frontend TypeScript check and production build


youchengsong (changes requested):
Blocking lifecycle findings:

1. In `_guard_native_goal_terminal` (around `backend/services/codex_app_server.py:1561-1609`), any `thread/goal/get` failure is converted to `goal = None`; when the Goal is between turns, `newer_turn_active` is false and CCM publishes a successful terminal event and detaches. This recreates the exact false-green/untracked-Goal failure this PR is meant to fix under a transient/invalid Goals RPC. Reproduction: emit active `thread/goal/updated`, then `turn/completed`, make `thread/goal/get` raise `CodexAppServerError`; `process.returncode` becomes 0 and the context disappears. Known-active/unknown state needs to retain exact generation evidence and retry/reconcile; success should require terminal proof.

2. `thread/goal/updated` and `thread/goal/cleared` only update `_thread_runtime` (around lines 3948-3957). Once the guard has observed active, it clears `pending_goal_terminal_notification`; a later complete/cleared notification between turns therefore leaves the retained process open forever. Reproduction: let the first guard return active, then emit `thread/goal/updated` with `status=complete`; the process remains pending and `_contexts_by_thread` still owns it. Retain enough terminal snapshot state, or explicitly reconcile/close retained idle contexts on terminal Goal notifications.

The five new tests pass, as do 23 Goal/steer-focused tests, but both adversarial regressions fail deterministically and need coverage before merge.

youchengsong (approved):
Fixed the two blocking retained-Goal lifecycle paths plus the Goal/descendant boundary race and cached-descendant adoption. Goal/steer/descendant focused tests pass (29), and the broader app-server + InstanceManager selection passes (187 tests). Latest main is merged and the worktree is clean.

PR #92, opened 2026-08-03
fix(chat): keep mobile composer above software keyboard

zxiangx:
## Summary

- synchronize the full-screen task chat with `window.visualViewport`
- keep the composer above the software keyboard in iOS in-app browsers such as Feishu WKWebView
- preserve the existing body scroll lock and leave desktop inline chat unchanged
- fall back to the current fixed layout when `visualViewport` is unavailable

## Root cause

The full-screen ChatView is fixed to the layout viewport while the iOS software keyboard only shrinks the visual viewport. Because the page body is also fixed to prevent scroll bleed, WebKit cannot reliably scroll the composer into view, leaving it behind the keyboard.

## Testing

- 120 focused chat and viewport tests pass
- `npm run build` passes (TypeScript + production Vite build)
- regression coverage includes resize, visual viewport scroll, orientation change, cleanup, fallback, and desktop inline mode

### DIFF
--- a/docs/plans/2026-08-03-ios-inapp-keyboard-design.md
+++ b/docs/plans/2026-08-03-ios-inapp-keyboard-design.md
@@ -1,0 +1,46 @@
+# iOS In-App Keyboard Viewport Design
+
+## Problem
+
+The full-screen task chat uses a fixed overlay and also locks `body` with
+`position: fixed`. In iOS Feishu's WKWebView, opening the software keyboard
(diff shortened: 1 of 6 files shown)


youchengsong (approved):
Reviewed the visualViewport bounds synchronization and its full-screen-only integration. Focused hook + ChatView suites pass (98 tests), and the production TypeScript/Vite build passes. No blocking findings.

PR #93, opened 2026-08-03
fix: scope task artifacts to project directories

luka7620:
## Summary

- inject one provider-neutral artifact contract into ordinary, Goal, Loop, follow-up, retry, resume, and Ralph turns
- require explicit user deliverables to live under the current Project's `.claude-manager/artifacts/task-<id>/` directory and emit links marked with the `ccm-task-artifact` Markdown title
- keep ordinary bare filename links such as `[DEPLOYMENT.md](DEPLOYMENT.md)` as normal links while retaining compatibility for legacy absolute paths and relative paths containing directories
- reject managed artifact paths belonging to a different Task before opening the file
- document the new convention and add backend/frontend regression coverage

## Why

The chat renderer previously treated every local-looking Markdown link as a Task artifact. Source references such as `[DEPLOYMENT.md](DEPLOYMENT.md)` therefore displayed a download icon and failed with `Artifact file not found`. At the same time, agents could leave generated deliverables outside a stable Project directory, so the Manager could not reliably serve them after a turn or temporary worktree ended.

## Compatibility and safety

- legacy Project file downloads outside the managed artifact namespace continue to work
- paths below `.claude-manager/artifacts/` must match the current `task-<id>` directory
- Tasks without a verifiable absolute `target_repo` are instructed not to emit downloadable local-file links
- artifact directories are kept out of commits through repository-local Git exclude rather than shared `.gitignore` changes

## Tests

- relevant backend suite: 275 passed, 1 deselected (`test_shutdown_cancels_and_reaps_auxiliary_lifecycles`, an unrelated current-upstream flaky test that also fails in isolation)
- `backend/tests/test_api_task_artifacts.py`: 16 passed
- Chat/Loop frontend suites: 113 passed
- TypeScript typecheck and `npm run build`
- targeted ESLint for the modified artifact/Loop modules
- `git diff --check`

## Local API end-to-end verification

- created a real Codex Task through `POST /api/tasks`
- Codex wrote a 2,476-byte UTF-8 artifact directly to `.claude-manager/artifacts/task-1/test-download.txt`, verified it was a regular non-symlink file, and returned a marked download link
- `GET /api/tasks/1/artifacts/download` returned `200 OK` with `Content-Disposition: attachment`
- source and downloaded files had identical SHA-256: `b43475eff84475f8e78d7b93c7aa94bc1d32e06b42bb789715efa313d9be2531`
- attempting to download a `task-2` managed path through Task 1 returned `403 Artifact belongs to a different task`


youchengsong (approved):
复审通过（修复后）：保留 project-scoped artifact 设计，并补齐三处阻塞边界——用户 prompt 伪造 policy tag 不再抑制权威前导；提示与下载共用安全项目根契约并拒绝 /NUL//不存在或 symlink 根；Manager 代理前握手 ，旧 Worker 在文件请求前 fail closed。已合入 #89 最新 main。后端受影响完整文件 302 passed；前端全量 525 passed；production build、Python compile、diff check 通过。

PR #100, opened 2026-08-05
feat: add per-session attention tags

zxiangx:
## Summary

- add one optional, user-owned `attention_tag` per Task without changing the existing machine-owned `tags`
- preserve the tag through clone, native Codex fork, Worker forwarding/migration, and shared-task backfill
- show and edit the tag inline from Task cards and the Chat header, including a mobile-friendly editor
- add Alembic migration, API validation, UI regression tests, and documentation

## Behavior

- trims leading and trailing whitespace
- treats blank input as removing the tag
- limits tags to 80 characters
- clone/fork inherits the tag unless the caller explicitly overrides or clears it

## Verification

- 129 focused frontend tests passed
- 439 focused backend API, fork, Worker, migration, and Alembic tests passed
- Alembic reports a single `2f6c8a1d4e90` head
- focused changed-file ESLint passed

## Notes

The editor is inline rather than a popover so it remains usable in embedded mobile browsers when the software keyboard opens.


youchengsong (approved):
已完成代码与安全审查并修复阻断项：Alembic 单 head、Shared attention_tag 同步、保存后前端状态一致性、移动端长按误触拖拽。验证：后端 3462 passed（仅排除 6 个已在干净 main 复现的既有失败），前端 566 passed，生产构建通过，迁移定向 27 passed。

PR #101, opened 2026-08-05
fix: prefer idle Codex accounts for task routing

Vicky-0719:
## Summary

- expose a read-only snapshot of Codex homes blocked by an active exec, ephemeral turn, or maintenance
- exclude those homes when routing fresh Codex tasks
- migrate and rebind an idle existing thread to another available account instead of repeatedly queueing behind an unrelated active turn
- retain the existing per-home admission guard as the final race-safe launch check

## Safety

- existing resident sessions remain sticky while their account is idle
- migration uses the existing rollout copy, registry rebind, durable binding, and compensation path
- when no compatible idle account exists, the existing retry/fail-closed behavior remains unchanged

## Tests

```text
638 passed, 610 warnings in 77.25s
```

Command:

```bash
python -m pytest backend/tests/test_codex_pool.py backend/tests/test_resume_config_dir.py backend/tests/test_service_instance_manager.py backend/tests/test_service_dispatcher.py -q
```


youchengsong (approved):
已基于最新 main 完成审查与修复：补齐 CODEX_HOME runtime 准入、取消/残留进程 fail-closed 和短退避路由回归。全后端 3446 项（排除 6 个已在干净 main 复现的既有失败）及最新主干交叉矩阵 1039 项通过。

PR #133, opened 2026-08-15
fix: verify instance process identity across restarts

luka7620:
## Summary

- Persist an exact OS process identity beside `Instance.pid`: PID, kernel start time, and boot session ID.
- Use that identity during startup recovery, manual retry, terminal cleanup, Task deletion, and stale reverse-owner reconciliation.
- Fence every cleanup with the persisted identity so a concurrent same-PID generation cannot be detached.
- Add a reversible Alembic migration plus Linux and macOS identity readers.
- Keep legacy or unreadable identities fail-closed, and retain the existing `claude_pty` external-effect boundary that blocks unsafe automatic replay.

## Problem

Recovery previously relied on `os.kill(pid, 0)`. After a host restart or PID reuse, an unrelated process can answer to the same PID, causing CCM to report that an unmanaged process may still be running and permanently block retry/capacity cleanup.

The exact start identity and boot ID distinguish the original managed process from a reused PID without weakening duplicate-execution protections.

## Validation

- `287 passed, 1 skipped`: process identity, startup recovery, Instance API, Task queue, and Plan runtime receipt suites.
- `19 passed`: Task retry regression selection.
- `3 passed`: stale reverse-owner reconciliation selection.
- New migration upgrade/downgrade/re-upgrade round-trip passed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,6 +94,7 @@ claude-manager/
 │       ├── plan_agent_runner.py # 严格只读 Planner/Reviewer pipeline + exact process cleanup
 │       ├── plan_tasks.py        # Plan 上下文快照、repo 指纹、stale/附件校验
 │       ├── plan_service.py      # Version 状态机、输入/审批、执行 Task 物化与 Worker outcome 导入
+│       ├── process_identity.py  # 跨平台 PID/start/boot 身份读取与重启存活校验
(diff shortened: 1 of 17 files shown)


youchengsong:
代码审查通过，逻辑和实现都没问题。但当前与 main 分支有冲突（刚合并了 #134），请 rebase 解决冲突后我来合并。

PR #135, opened 2026-08-16
feat(codex): configure pool routing from the account UI

LianningLBY:
## Summary

- persist non-secret Codex pool runtime policy beside the private account config
- add admin GET/PUT settings endpoints and a Pro → Codex settings UI
- support pause/resume, cooldown, proactive quota threshold, API/OAuth routing order, and durable preferred account
- fail closed across main tasks, Monitor, Sub-Agent, and Distill while paused
- serialize settings/preferred writes through the Worker node account-mutation fence

Environment variables remain deployment-level controls for whether the native pool exists, its storage path, and legacy defaults. Saved runtime policy takes effect immediately and survives service restarts.

## Compatibility and safety

- legacy account JSON without `pool_settings` keeps existing defaults
- writes use the existing atomic private JSON path (`0600`) and restore exact prior bytes if activation cannot be verified
- deployment-level disabled status still returns a stable frontend response shape
- paused routing never falls back to ambient `CODEX_HOME`

## Validation

- `AUTH_TOKEN=task281-test ... pytest backend/tests/test_codex_pool.py backend/tests/test_skill_distill.py -q` — 95 passed
- Worker/Monitor/Resume matrix — 137 passed; one unrelated existing Claude relogin case could not locate `.venv` in the isolated worktree and returned 501
- `npx vitest run src/components/Layout/PoolDrawer.test.tsx` — 49 passed
- `npx tsc --noEmit` — passed
- `npm run build` — passed (4764 modules)
- `git diff --check` — passed

Ruff on the touched backend file set reports existing latest-main findings in `dispatcher.py`, `skill_distill.py`, and an existing unused `secrets` import in `codex_pool.py`; this change does not expand those baseline findings.


youchengsong (commented):

## Reviewer summaries

The persistence design is generally cohesive, but preferred-account validation is performed outside the mutation fence and local lock, allowing a durable dangling preference under concurrent account mutation.

The routing changes are coherent, but the settings persistence path does not uphold its stated rollback guarantee when activation fails.

设置持久化的成功路径有覆盖，但激活异常时无法兑现原字节回滚保证，可能造成磁盘与运行状态不一致。

backend/api/codex_pool.py
```diff
@@ -1529,6 +1542,70 @@ async def codex_pool_status():
...
+        data = json.loads(original_bytes.decode("utf-8"))
+        if not isinstance(data, dict):
+            raise RuntimeError("Invalid Codex pool config")
+    else:
+        data = {"accounts": []}
+    data["pool_settings"] = values
+    _write_private_json(pool_path, data)
+    pool.reload()
+    effective = pool.settings()
+    if effective != values:
```
youchengsong, line 1577: **[MEDIUM] Preferred-account validation occurs before mutation serialization**

Evidence: Both `codex_update_pool_settings` and `codex_set_preferred` call `pool.account(...)` before `await fence_worker_node_account_mutation(db)` and before entering `_login_lock`, then later persist the previously validated ID. The patch also contains account deletion/retirement operations sharing this mutable account state, so an account can change after validation but before the settings file is written.

Impact: A concurrent delete or retirement can leave `pool_settings.preferred_account_id` pointing to an unavailable account. The request can report success and persist inconsistent state even though the endpoint promises to reject unknown or retired preferences.

Required fix: After acquiring the worker mutation fence, acquire `_login_lock` and revalidate `preferred_account_id` against the current pool state immediately before `_persist_pool_settings`. Apply the same ordering to both settings and preferred endpoints.

Verification: Block a settings/preferred request between its initial state observation and lock acquisition, delete or retire the selected account, then resume the request. Assert it returns 404 and that neither the persisted `pool_settings.preferred_account_id` nor the effective pool preference references that account.

Finding fingerprint: `91a7104a20e25c03d6591c21f0d9dab2fc7bca1a1c2233e194b647d1b216b428`

backend/api/codex_pool.py
```diff
@@ -1529,6 +1542,70 @@ async def codex_pool_status():
...
+
+    pool_path = _pool_config_path(pool)
+    snapshot = _snapshot_private_file(pool_path)
+    if snapshot["existed"]:
+        original_bytes = base64.b64decode(snapshot["content_b64"], validate=True)
+        data = json.loads(original_bytes.decode("utf-8"))
+        if not isinstance(data, dict):
+            raise RuntimeError("Invalid Codex pool config")
+    else:
+        data = {"accounts": []}
```
youchengsong, line 1564: **[MEDIUM] Activation failures leave the newly written configuration in place**

Evidence: `_persist_pool_settings` snapshots the original file, calls `_write_private_json`, and then calls `pool.reload()` outside any rollback guard. Restoration only occurs after reload succeeds and `effective != values`. If `pool.reload()` raises, including the explicitly handled `RuntimeError`, `ValueError`, or `JSONDecodeError`, the function exits without restoring the snapshot. The PUT endpoint then reports failure while the new bytes remain on disk and may take effect on a later reload.

Impact: A failed settings request can durably change routing policy despite returning an error. This violates the patch's exact-byte restoration safety claim and can unexpectedly pause routing or alter account selection after restart.

Required fix: Wrap the write, reload, and verification sequence so every exception or verification failure after the write restores the snapshot and reloads the prior configuration before propagating the original failure. Preserve or report rollback failure explicitly if restoration itself fails.

Verification: Start with known config bytes, make the first `pool.reload()` raise after `_write_private_json`, invoke `_persist_pool_settings`, and assert it raises while the file contents are byte-for-byte identical to the snapshot and the pool is reloaded from the restored configuration.

Finding fingerprint: `df5c3e0c81c5ddbe4fcc6a94b90b7bd35e1f628e6014a355327d1d38af671a3d`

backend/api/codex_pool.py
youchengsong, line 1577: **[MEDIUM] reload 抛错时新配置不会回滚**

Evidence: `_persist_pool_settings` 先执行 `_write_private_json`，随后直接调用 `pool.reload()`；仅在 reload 成功且 `effective != values` 时才恢复 snapshot。若 `pool.reload()` 因 I/O、配置加载或账号解析异常而抛错，外层端点会返回 500，但没有恢复已经写入的文件。

Impact: 请求报告保存失败时，新策略仍可能留在磁盘，而进程内 pool 可能保持旧状态或部分加载状态；重启后失败的启停或路由设置还可能突然生效，违反 PR 声明的“激活无法验证时恢复精确原字节”。

Required fix: 将写入后的 reload、effective 读取和比较置于统一的异常回滚路径；任何激活异常或不匹配都应恢复 snapshot，并重新加载原配置后再向上抛错。

Verification: 新增失败路径测试：保存原文件字节，令首次 `pool.reload()` 抛出异常、恢复后的 reload 成功；断言 API/辅助函数失败、文件字节与原值完全相同，并且 pool 的有效设置恢复为原配置。

Finding fingerprint: `309f6b3a4e06987a3f37928619d47312d986b4fbafaf20500e073d4ff3f45f3f`

PR #136, opened 2026-08-17
fix(api-accounts): support Apex Claude models

LianningLBY:
## Summary

- support both native `models[].slug` and OpenAI-compatible `data[].id` Apex model catalogs
- project Apex Claude and Codex models into their respective account pools
- generate Anthropic Messages and OpenAI Responses runtime configurations from the same Apex key
- safely migrate legacy Codex-only Apex account metadata while rejecting modified Claude runtime files
- update the account drawer copy and regression coverage

## Validation

- `uv run pytest -q backend/tests/test_cloudrouter_accounts.py backend/tests/test_claude_pool.py backend/tests/test_codex_pool.py` — 354 passed
- `vitest run src/components/Layout/PoolDrawer.test.tsx` — 49 passed
- TypeScript build and Vite production build passed
- manually verified on the isolated test service at port 8010: Apex-backed Claude Opus 4.6 task completed successfully


youchengsong (commented):

## Reviewer summaries

The dual-provider projection reuses the existing account abstraction, but the legacy migration mutates managed state before completing its fail-closed validation.

### Senior engineer — Passed

变更正确支持 Apex 双模型目录格式和双提供商运行时，并对旧配置执行严格、可恢复的迁移；未发现具体阻塞问题。

Finding count: none.

### QA engineer — Passed

The patch delivers dual-provider Apex model discovery and runtime configuration, includes fail-closed legacy migration behavior, and adds meaningful backend and frontend regression coverage. No blocking QA issue is evident from the supplied material.

Finding count: none.

backend/services/cloudrouter_accounts.py
```diff
@@ -1880,6 +1906,8 @@ def _load_account(self, path: Path) -> CloudRouterAccount:
                 ("account.json", 0o600), ("api.key", 0o600), ("key-helper", 0o700),
             ):
                 _require_owned_regular(path / file_name, expected_mode)
+            if migrate_apex_claude_runtime:
```
youchengsong, line 1910: **[MEDIUM] Legacy migration writes files before validating the existing account**

Reviewer: `principal_engineer` · Category: `operations`

Evidence: `_load_account` invokes `_migrate_apex_claude_runtime(account)`, which creates missing Claude files, before `_validate_runtime_configuration(account)`. If validation then rejects an existing modified Codex configuration or another managed artifact, loading fails after the account has already been changed; the endpoint metadata update occurs even later.

Impact: Merely attempting to load an unsafe legacy account can partially migrate it, weakening fail-closed behavior and complicating diagnosis, rollback, and administrative recovery. A failed validation should not modify managed account state.

Required fix: Preflight all existing legacy managed files, including Codex configuration and any existing Claude files, before creating migration artifacts. Only after that validation succeeds should the idempotent Claude-file and endpoint-metadata migration begin.

Verification: Create a legacy Apex account with no Claude runtime files and a modified Codex managed configuration, call `reload()`, assert the unsafe-path error, and assert that both Claude files remain absent and the endpoint metadata remains the legacy snapshot.

Finding fingerprint: `28e268279473ff6dfed0f70544490d872bfe4d1a7b4d336c97c3dcfd8f3131a4`

luka7620: Fixed in 346a763d. Legacy Apex loading now performs a read-only preflight of every existing Claude artifact and the Codex runtime before materializing missing Claude files or updating endpoint metadata. Added test_legacy_apex_migration_preflights_codex_before_writing, which corrupts the managed Codex route and verifies both Claude files remain absent and metadata bytes remain unchanged. Validation: 103 backend account tests passed; both exact-head CI checks passed.

PR #138, opened 2026-08-24
fix(claude): recover stalled response loops

LianningLBY:
## Summary

- detect sustained, similar Claude assistant responses whose `stop_reason` remains null and stop the exact foreground PTY generation
- clear the unhealthy native session and perform one generation-bound fresh-session retry only when the turn had no tool activity
- fail closed after tool activity, on a second stalled retry, or when the volatile retry permit is missing or stale
- preserve Claude `stop_reason` in parsed stream events and cover the Task 584 event shape with regression tests

## Safety

- any `tool_use` or `tool_result` disables automatic replay for that turn
- recovery advances the task generation and never resumes the unhealthy session
- late events from an old PTY session cannot interrupt a newer session
- automatic recovery is limited to one attempt

## Verification

- focused no-progress regression: `10 passed`
- affected suites: `1028 passed, 1 failed`; the remaining CloudRouter PTY binary projection failure also reproduces on unchanged `main`
- local CCM synthetic Task 584 reproduction triggered the guard, interrupted the PTY once, cleared the session, and completed a fresh-session recovery
- real Claude API smoke test completed with `stop_reason=end_turn`, no tool activity, no retry, and no false-positive guard trigger

This PR only adds the guard and recovery behavior; it does not change the upstream Claude gateway.

### DIFF
--- a/backend/services/dispatcher.py
+++ b/backend/services/dispatcher.py
@@ -770,7 +770,7 @@ class QueuedMessage:
     claimed_turn_generation: int | None = field(
         compare=False, default=None, repr=False
     )
-    # Once a compact retry has claimed G+1, bind later in-process launch
+    # Once an automatic retry has claimed G+1, bind later in-process launch
     # retries to the exact source row installed by that claim.  The original
     # permit still names rejected G; this field prevents it from floating to a
     # different G+1 source after a pre-launch rollback/requeue.
@@ -811,6 +811,12 @@ class QueuedMessage:
     context_retry_permit: "ContextRetryPermit | None" = field(
         compare=False,
         default=None,
+        repr=False,
+    )
+    # A no-progress recovery may replay the exact user input only once.
+    no_progress_retry_attempt: int = field(
+        compare=False,
+        default=0,
         repr=False,
     )
     # Internal producers snapshot this before committing the result which
@@ -857,17 +863,17 @@ def _context_retry_permit_matches(
     task: Task,
     msg: QueuedMessage,
 ) -> bool:
-    """Validate the exact rejected generation before compact-retry admission."""
+    """Validate the exact rejected generation before automatic retry."""
 
     permit = msg.context_retry_permit
     if permit is None:
-        # ``compact_retry`` is an internal replay authority, not merely a
-        # descriptive source label.  Legacy/forged queue items without the
-        # immutable rejected-generation proof must fail closed; ordinary
-        # queued messages do not need this specialized permit.
-        return msg.source != "compact_retry"
-    if msg.source != "compact_retry" or not isinstance(
-        permit, ContextRetryPermit
+        # Internal replay sources are authorities, not merely descriptive
+        # labels. Legacy/forged queue items without immutable generation proof
+        # must fail closed; ordinary messages do not need this permit.
+        return msg.source not in {"compact_retry", "no_progress_retry"}
+    if (
+        msg.source not in {"compact_retry", "no_progress_retry"}
+        or not isinstance(permit, ContextRetryPermit)
     ):
         return False
     baseline_matches = bool(
@@ -1291,7 +1297,7 @@ class GlobalDispatcher:
         started_at: datetime | None,
         completed_at: datetime | None,
     ) -> ContextRetryPermit:
-        """Issue one volatile authority for rejected G -> compact G+1."""
+        """Issue one volatile authority for a proven-safe rejected G -> G+1."""
 
         authority_id = secrets.token_hex(16)
         permit = ContextRetryPermit(
@@ -1305,7 +1311,7 @@ class GlobalDispatcher:
             started_at=started_at,
             completed_at=completed_at,
         )
-        # Only one compact authority for an exact rejected generation may be
+        # Only one retry authority for an exact rejected generation may be
         # live in this process.  Re-issuing revokes an older unconsumed copy.
         for existing_id, existing in tuple(
             self._context_retry_authorities.items()
@@ -1320,7 +1326,7 @@ class GlobalDispatcher:
         return permit
 
     def _context_retry_authority_is_live(self, msg: QueuedMessage) -> bool:
-        if msg.source != "compact_retry":
+        if msg.source not in {"compact_retry", "no_progress_retry"}:
             return msg.context_retry_permit is None
         permit = msg.context_retry_permit
         return bool(
@@ -1345,11 +1351,11 @@ class GlobalDispatcher:
             self._context_retry_authorities.pop(authority_id, None)
 
     def _consume_context_retry_authority(self, msg: QueuedMessage) -> None:
-        if msg.source != "compact_retry":
+        if msg.source not in {"compact_retry", "no_progress_retry"}:
             return
         if not self._context_retry_authority_is_live(msg):
             raise QueuedMessagePrelaunchError(
-                "Compact retry launch authority is no longer current"
+                "Automatic retry launch authority is no longer current"
             )
         self.revoke_context_retry_permit(msg.context_retry_permit)
 
@@ -18718,6 +18724,7 @@ Codex 中工具会显示为上述 mcp__ccm_monitor_agent__* canonical 名称；
         queue_timestamp: float | None = None,
         allow_new_session: bool | None = None,
         context_retry_permit: ContextRetryPermit | None = None,
+        no_progress_retry_attempt: int = 0,
         queue_admission_fence: QueueAdmissionFence | None = None,
         initiating_user_id: int | None = None,
         initiating_user_role: str = "member",
@@ -18745,6 +18752,19 @@ Codex 中工具会显示为上述 mcp__ccm_monitor_agent__* canonical 名称；
         ):
             raise ValueError(
                 "Shared messages must use the sandboxed system principal"
+            )
+        if source == "no_progress_retry":
+            if (
+                no_progress_retry_attempt != 1
+                or context_retry_permit is None
+                or allow_new_session is not True
+            ):
+                raise ValueError(
+                    "No-progress retry requires one fresh-session authority"
+                )
+        elif no_progress_retry_attempt != 0:
+            raise ValueError(
+                "No-progress retry attempt is only valid for its internal source"
             )
         internal_session_report = source.startswith(("monitor:", "sub-agent:"))
         msg = QueuedMessage(
@@ -18775,6 +18795,7 @@ Codex 中工具会显示为上述 mcp__ccm_monitor_agent__* canonical 名称；
                 allow_new_session is None and internal_session_report
             ),
             context_retry_permit=context_retry_permit,
+            no_progress_retry_attempt=no_progress_retry_attempt,
             queue_admission_fence=queue_admission_fence,
         )
         async with self._dispatch_claim_lock:
@@ -23726,6 +23747,7 @@ Codex 中工具会显示为上述 mcp__ccm_monitor_agent__* canonical 名称；
                 ssh_agent_socket_snapshot=msg.ssh_agent_socket_snapshot,
                 attachment_paths=msg.attachment_paths,
                 context_retry_permit=msg.context_retry_permit,
+                no_progress_retry_attempt=msg.no_progress_retry_attempt,
                 context_retry_claimed_source_log_id=(
                     msg.context_retry_claimed_source_log_id
                 ),
--- a/backend/services/dispatcher.py
+++ b/backend/services/dispatcher.py
@@ -1350,6 +1350,98 @@ class GlobalDispatcher:
         ):
             self._context_retry_authorities.pop(authority_id, None)
 
+    async def enqueue_no_progress_recovery(
+        self, *, task_id: int, instance_id: int, record, params: dict
+    ) -> bool:
+        """Authorize and enqueue one exact-generation Claude recovery.
+
+        InstanceManager supplies only the observed record and launch metadata;
+        all durable source validation and retry authority remain Dispatcher
+        policy, avoiding a dependency on the application singleton.
+        """
+        from backend.services.terminal_arbitration import (
+            source_alias_original_log_id,
+            source_shape_is_canonical,
+        )
+
+        requested_source_id = params.get("source_log_id")
+        async with self.db_factory() as db:
+            task = await db.get(Task, task_id)
+            if (
+                task is None
+                or (task.provider or "claude").lower() != "claude"
+                or task.status != "failed"
+                or task.instance_id != instance_id
+                or task.retry_count != record.task_retry_count
+                or task.turn_generation != record.task_turn_generation
+                or task.session_id is not None
+                or type(task.turn_source_log_id) is not int
+                or type(requested_source_id) is not int
+                or requested_source_id <= 0
+                or task.turn_source_log_id <= 0
+            ):
+                return False
+            source = await db.get(LogEntry, task.turn_source_log_id)
+            original_id = source_alias_original_log_id(source) if source else None
+            original = await db.get(LogEntry, original_id) if original_id else None
+            if (
+                source is None
+                or source.task_id != task.id
+                or source.task_retry_count != task.retry_count
+                or source.task_turn_generation != task.turn_generation
+                or source.turn_scope != "source"
+                or requested_source_id not in {source.id, original_id}
+                or not source_shape_is_canonical(source, original)
+            ):
+                return False
+            replay_source = original if source.event_type == "turn_source" else source
+            prompt = replay_source.content if replay_source else None
+            if not isinstance(prompt, str) or not prompt:
+                return False
+            permit = self.issue_context_retry_permit(
+                task_id=task.id,
+                instance_id=instance_id,
+                retry_count=task.retry_count,
+                turn_generation=task.turn_generation,
+                turn_source_log_id=task.turn_source_log_id,
+                session_id=None,
+                started_at=task.started_at,
+                completed_at=task.completed_at,
+            )
+        retry_kwargs = {
+            "task_id": task_id,
+            "prompt": prompt,
+            "priority": PRIORITY_USER,
+            "source": "no_progress_retry",
+            "source_log_id": requested_source_id,
+            "current_message": prompt,
+            "allow_new_session": True,
+            "context_retry_permit": permit,
+            "no_progress_retry_attempt": 1,
+            "initiating_user_id": params.get("initiating_user_id"),
+            "initiating_user_role": params.get("initiating_user_role", "member"),
+            "execution_mode": params.get("execution_mode", "sandbox"),
+            "execution_principal_kind": params.get("execution_principal_kind", "system"),
+            "attachment_paths": tuple(params.get("attachment_paths") or ()),
+            "ssh_agent_socket_snapshot": params.get("ssh_agent_socket_snapshot"),
+        }
+        if isinstance(params.get("enabled_skills"), dict):
+            retry_kwargs["command_skills"] = dict(params["enabled_skills"])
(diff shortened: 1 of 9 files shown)


youchengsong (commented):

The panel found 4 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

Recovery is implemented in two terminal-finalization paths, allowing duplicate replay queue entries and revocation of a still-pending retry authority.

The recovery path lacks proof that it replays the authoritative user input, and the parser conflates an absent stop_reason with an explicit null value, enabling false recovery triggers.

The recovery path is substantially tested, but tool activity in the same Claude assistant envelope can be discovered too late to prevent an erroneous interruption.

backend/services/instance_manager.py
```diff
@@ -12263,6 +12387,18 @@ async def withdraw_admission_for_background_handoff(db) -> bool:
                             return None
```
youchengsong, line 12412: **[MEDIUM] No-progress recovery is enqueued from two terminal paths**

Evidence: The patch calls `_try_enqueue_no_progress_recovery(instance_id, task_id, record)` in `withdraw_admission_for_background_handoff` and calls the same method again later in `_consume_output_impl` around line 14222. Both calls can observe the same failed task with `session_id is None` before the queued retry is processed; issuing the second permit revokes the first permit while its message is still in the queue.

Impact: A single stalled turn can produce duplicate recovery queue entries. The first entry may later fail prelaunch with a revoked authority, creating spurious task/dispatcher failures and making recovery behavior dependent on queue-worker timing, contrary to the documented one-attempt guarantee.

Required fix: Choose one authoritative terminal-finalization location for no-progress recovery, or add an atomic durable/idempotent claim that both call sites must acquire before issuing a permit and enqueueing the replay.

Verification: Add a regression test that invokes both terminal paths for the same failed generation before queue processing and asserts exactly one permit issuance and one queued `no_progress_retry` message, with no revoked-authority prelaunch failure.

Finding fingerprint: `25580dfe9745ab8e0f8c6f460902c7590efaff962235681ed3515dc24c662206`

backend/services/instance_manager.py
```diff
@@ -16632,6 +16970,96 @@ async def guard_managed_event_generation(db) -> bool:
                 task_id,
```
youchengsong, line 16903: **[HIGH] Text is evaluated before sibling tool activity in the same assistant envelope**

Evidence: The stream parser can emit multiple events from one assistant message and assigns the shared stop_reason to every event. The no-progress state, however, learns about tool activity only when observe() receives an event whose event_type is tool_use/tool_result. If an assistant envelope contains a text block followed by a tool_use block, its text event can satisfy the three-message/two-minute threshold and immediately call _interrupt_no_progress_claude_turn before the sibling tool_use event is processed. Added tests cover tool activity arriving before subsequent messages, but not mixed text/tool assistant envelopes.

Impact: A legitimate Claude turn that has already selected a tool can be interrupted, marked failed, and have its native session cleared. Although the later tool event may disable replay, it cannot undo the erroneous interruption, violating the requirement that recovery apply only to turns with no tool activity.

Required fix: Make tool presence known before evaluating any text event from the same parsed envelope—for example, pre-scan each parsed event batch or annotate all sibling events when the raw assistant message contains a tool_use block—and suppress both interruption and replay for that turn.

Verification: Feed the production parser/consumer three sufficiently similar incomplete assistant messages over the threshold, with the triggering assistant envelope containing text followed by tool_use. Assert no PTY interrupt occurs, no no-progress fatal error is set, and no automatic recovery is enqueued.

Finding fingerprint: `b1c47ab980a497e1c6cfccbd9f6d9218a08c761311e46eee415c0e2bc78f1433`

backend/services/instance_manager.py
```diff
@@ -14798,6 +14985,148 @@ async def _chat_automatic_relaunch_is_blocked(
...
+    async def _try_enqueue_no_progress_recovery(
+        self,
+        instance_id: int,
+        task_id: int,
+        record: _OutputConsumerRecord,
+    ) -> bool:
+        """Queue one fresh-session replay after a proven effect-free loop."""
+
+        state = record.claude_no_progress
+        params = self._launch_params.get(instance_id) or {}
```
youchengsong, line 15022: **[MEDIUM] Automatic recovery does not verify the replayed prompt against the source log**

Evidence: _try_enqueue_no_progress_recovery chooses `params.get("current_message") or params.get("prompt")`. It validates the identity and canonical shape of the source LogEntry but never compares this volatile prompt with the authoritative source/original content. The fallback can therefore enqueue a wrapped, stale, or otherwise different launch prompt while claiming to replay the exact user input.

Impact: A no-progress failure can automatically execute instructions other than the original user message on the fresh session. The generation permit proves which turn may retry, but not what content may be replayed, so the advertised replay safety boundary is incomplete.

Required fix: Derive the retry text from the validated canonical source entry, or explicitly verify that `current_message` exactly matches its authoritative raw user content. Remove the fallback to the potentially wrapped `prompt` and fail closed on any mismatch or missing canonical content.

Verification: Create a canonical source whose user content differs from both `current_message` and `prompt`; assert that no retry is enqueued. Also cover a missing `current_message` with a wrapped `prompt` and assert fail-closed behavior, while retaining a success case for an exact match.

Finding fingerprint: `0480176e40589a1c5cad4d29c937b15d8a0fee5974d68e6d00d971f409a08c08`

backend/services/stream_parser.py
```diff
@@ -168,6 +168,7 @@ def _base_event() -> dict:
                 event = _base_event()
                 event["role"] = "assistant"
                 event["event_type"] = "message"
+                event["stop_reason"] = message_obj.get("stop_reason")
```
youchengsong: **[MEDIUM] Missing stop_reason is incorrectly represented as explicit null**

Evidence: The parser unconditionally assigns `event["stop_reason"] = message_obj.get("stop_reason")`, including in the loop over all parsed events. For an assistant envelope where the field is absent, this synthesizes a null key. `_ClaudeNoProgressState._stop_reason` treats the presence of that key as proof that stop_reason was supplied, so such events count as incomplete responses rather than unknown metadata.

Impact: Three similar legacy or variant assistant events that omit stop_reason can trigger an interrupt and automatic replay after two minutes, even though the required explicit-null signal was never observed.

Required fix: Only attach `stop_reason` when `"stop_reason" in message_obj`; preserve the distinction between an absent field and an explicitly null field in every parser branch.

Verification: Parse three assistant envelopes that omit stop_reason and feed them to _ClaudeNoProgressState across more than 120 seconds; assert that no messages are counted and no trigger occurs. Retain separate tests proving that explicit null counts and `end_turn` resets detection.

Finding fingerprint: `5ed5f04debf82e661ed020e1d0ca3ad0f07ae8fde9f10c8dee8c9b734c214778`

backend/services/instance_manager.py
LianningLBY: Fixed in commit 8b664a6f. Recovery no longer uses current_message or falls back to the wrapped launch prompt. After canonical source/provenance validation it derives replay text from the authoritative user LogEntry (the original row for an alias) and fails closed if no non-empty user content exists. The test supplies a stale launch parameter and verifies only the source-log content is enqueued.

backend/services/stream_parser.py
LianningLBY: Fixed in commit 8b664a6f. StreamParser now adds stop_reason only when the upstream assistant message actually contains the field, preserving absent versus explicit null semantics in every assistant branch. Tests cover both parser shapes and prove missing metadata never increments or triggers the no-progress state.

backend/services/instance_manager.py
LianningLBY: Fixed in commit 8b664a6f. The no-progress state now claims recovery synchronously before either terminal path can await, so the shared exact-generation consumer record can issue only one permit and enqueue only one no_progress_retry. The regression calls the recovery entry twice before queue processing and asserts one permit and one enqueue.

backend/services/instance_manager.py
LianningLBY: Fixed in commit 8b664a6f. The parser pre-scans each assistant envelope for tool_use and annotates every sibling event. The no-progress state consumes that annotation before evaluating a text event, so text-before-tool ordering cannot trigger an interrupt or replay. Regression coverage verifies the mixed text/tool envelope and state transition.

youchengsong (commented):

## Reviewer summaries

The recovery design reuses existing generation-bound retry authority, but its detector introduces unbounded synchronous work into the event-processing path.

The recovery flow is carefully fenced, but its similarity detector introduces unbounded synchronous work on provider-controlled output.

### QA engineer — Passed

The patch implements generation-bound, single-attempt recovery for sustained Claude no-progress loops, fails closed after tool activity or stale authority, clears unhealthy sessions, and includes meaningful detector, parser, interruption, terminal-state, duplicate-enqueue, and dispatcher-admission coverage.

Finding count: none.

backend/services/instance_manager.py
```diff
@@ -759,6 +760,124 @@ class _OutputConsumerRecord:
...
+        if not isinstance(content, str):
+            return False
+        normalized = self._normalize(content)
+        if len(normalized) < 5:
+            return False
+
+        if self.anchor_text is None:
+            self.anchor_text = normalized
+            self.similar_messages = 1
+            self.first_seen_monotonic = now
```
youchengsong, line 873: **[MEDIUM] Unbounded similarity comparison can block the async event loop**

Reviewer: `senior_engineer` · Category: `performance`

Evidence: _ClaudeNoProgressState.observe passes the complete normalized anchor and current assistant response to difflib.SequenceMatcher(..., autojunk=False).ratio() without any length bound. SequenceMatcher has quadratic or worse pathological behavior, and disabling autojunk removes its repeated-token mitigation. This runs synchronously in _process_event.

Impact: A sufficiently long or repetition-heavy Claude response can monopolize the service event loop, delaying all instance processing and potentially recreating a service-wide stall while trying to detect a per-turn stall.

Required fix: Bound the amount of text processed by the detector or replace SequenceMatcher with a similarity calculation whose time and memory are explicitly bounded; preserve representative text rather than comparing unrestricted provider output.

Verification: Feed very large, repetition-heavy assistant messages through observe and verify structurally—such as by spying on the comparison inputs or exercising the bounded helper—that work and retained state never exceed the configured limit, while the existing Task 584 sequence still triggers.

Finding fingerprint: `ae9722bde4316f0ac257e8256f923ad99a7218175fa269aca4f9dfc18ed4733c`

backend/services/instance_manager.py
```diff
@@ -759,6 +760,124 @@ class _OutputConsumerRecord:
...
+            return False
+
+        similarity = difflib.SequenceMatcher(
+            None,
+            self.anchor_text,
+            normalized,
+            autojunk=False,
+        ).ratio()
```
youchengsong, line 881: **[MEDIUM] Unbounded SequenceMatcher can stall the shared event loop**

Reviewer: `principal_engineer` · Category: `performance`

Evidence: _ClaudeNoProgressState.observe stores the complete provider-controlled assistant content and synchronously runs difflib.SequenceMatcher(..., autojunk=False).ratio() without any input-length bound. This method is called directly from the asynchronous _process_event path, and SequenceMatcher has quadratic worst-case behavior, particularly for repetitive inputs when autojunk is disabled.

Impact: A large or pathological Claude response can monopolize the asyncio event loop, delaying interrupts, terminal finalization, watchdogs, and unrelated instance processing. This is especially hazardous in the unhealthy-response path the guard is intended to recover.

Required fix: Bound the text retained and compared, or replace SequenceMatcher with a bounded-cost similarity calculation. Preserve a representative bounded sample if both prefixes and suffixes matter, and document the maximum work per event.

Verification: Add a deterministic regression that feeds very large repetitive assistant messages and verifies the similarity implementation never receives or retains more than the configured bound, while three bounded similar observations still trigger recovery after the elapsed threshold.

Finding fingerprint: `e6ca74e0bffe4873d371005efbe628eb84720b4c5ea0d07fc4029028ffa9b609`

backend/services/instance_manager.py
LianningLBY: Fixed in commit cdf2e7f8. The detector now samples at most 1024 characters before normalization, retaining both the prefix and suffix, and applies the same cap again after Unicode case folding. Stored anchor text and both SequenceMatcher inputs therefore have a hard 1024-character limit. Regression coverage feeds more than one million repetition-heavy characters, spies on every comparison input length, verifies the retained state bound, and confirms the loop still triggers after 121 seconds.

backend/services/instance_manager.py
LianningLBY: Fixed in commit cdf2e7f8. Similarity work is now explicitly bounded to 1024 characters per side using a representative prefix-plus-suffix sample before normalization and a second post-casefold cap. A deterministic large repetitive-input test records the actual SequenceMatcher input sizes and asserts both sides and retained anchor state stay within the limit, while the existing Task 584 sequence remains green.

youchengsong (commented):

## Reviewer summaries

The recovery policy creates a bidirectional dependency between dispatcher and instance lifecycle layers through the application singleton.

### Senior engineer — Passed

The stalled-response detection, generation-bound retry authorization, tool-activity fail-closed behavior, exact-process interruption, and one-attempt recovery flow are coherent and covered by focused regression tests. No concrete blocking defect was identified in the supplied patch.

Finding count: none.

The detection and retry components are well covered individually, but the safety-critical tool-activity replay prohibition is not proven through the production workflow.

backend/tests/test_service_instance_manager.py
```diff
@@ -162,6 +164,429 @@ def test_claude_terminal_result_suppresses_only_exact_assistant_duplicate():
...
+
+def test_claude_no_progress_state_ignores_real_progress_and_terminal_answers():
+    state = _ClaudeNoProgressState()
+
+    assert not state.observe(
+        _incomplete_claude_message("I will inspect the backend logs now."),
+        now=0,
+    )
+    assert not state.observe(
+        _incomplete_claude_message("The database query found 18 rows."),
```
youchengsong, line 245: **[MEDIUM] Tool-result replay prohibition lacks production-path coverage**

Evidence: The safety claim says any tool_use or tool_result disables automatic replay. The state unit test injects only a hand-built tool_use event, the parser test covers only a mixed tool_use envelope, the recovery test manually sets state fields and mocks dispatcher.enqueue_message, and the process-event test contains no tool activity or terminal recovery. No supplied test sends a parsed tool_result through _process_event and terminal finalization and verifies that no no_progress_retry is enqueued.

Impact: A parser, event-association, or finalization integration regression could replay a turn after a tool side effect, duplicating external actions despite the PR's primary fail-closed guarantee.

Required fix: Add one focused production-path regression that processes real parsed tool_result activity on the tracked Claude generation, subsequently triggers the stalled-response condition, completes terminal handling, and asserts no no_progress_retry admission occurs. Cover tool_use similarly if the same path is not shared.

Verification: Drive StreamParser output through InstanceManager._process_event and the relevant PTY terminal finalizer with a dispatcher spy; assert tool_activity_seen is true and dispatcher.enqueue_message is never awaited with source='no_progress_retry'.

Finding fingerprint: `99d621efebe5aa4e39f7ba346572be090157342e3c436ed9f020c4c15991edd4`

backend/services/instance_manager.py
```diff
@@ -14798,6 +15008,162 @@ async def _chat_automatic_relaunch_is_blocked(
...
+        """Queue one fresh-session replay after a proven effect-free loop."""
+
+        state = record.claude_no_progress
+        params = self._launch_params.get(instance_id) or {}
+        attempt = params.get("no_progress_retry_attempt", 0)
+        requested_source_id = params.get("source_log_id")
+        if not (
+            state.triggered
+            and not state.tool_activity_seen
+            and not state.recovery_claimed
```
youchengsong, line 15028: **[MEDIUM] Recovery reaches back into the dispatcher composition root**

Reviewer: `principal_engineer` · Category: `architecture`

Evidence: Dispatcher already owns and launches InstanceManager, while `_try_enqueue_no_progress_recovery` dynamically imports `backend.main.dispatcher`, issues its permit, and calls `enqueue_message`. The new tests must patch `backend.main.dispatcher` to exercise this path. This places queue authorization and replay orchestration inside the subprocess lifecycle service and creates a dispatcher -> instance manager -> application singleton/dispatcher dependency cycle.

Impact: InstanceManager behavior now depends on global application initialization and cannot be reused or tested through its declared collaborators. Alternate workers or initialization paths can silently fail recovery when the singleton is unavailable, and retry admission policy has two ownership points across modules.

Required fix: Introduce a narrow injected recovery-scheduling capability (or return a typed recovery request to the dispatcher). Keep loop detection and interruption in InstanceManager, but have the dispatcher-owned capability validate the generation/source, issue the permit, and enqueue the replay without importing `backend.main`.

Verification: Construct InstanceManager with a fake recovery scheduler, finalize a qualifying failed generation, and assert the typed request is emitted once. Separately test the dispatcher scheduler against a fake manager/database for exact-generation admission, stale-generation rejection, duplicate suppression, and enqueue failure, without patching `backend.main.dispatcher`.

Finding fingerprint: `e8b1c0a608d37c377e87f4d49857f2ba489d8303f0e692dde004be8cefe4fcb1`

youchengsong (commented):

## Reviewer summaries

The recovery capability is placed behind a Dispatcher-owned scheduler, reuses existing generation-bound retry authority, validates canonical turn sources, and includes concurrency and tool-activity safeguards. No concrete blocking architectural or behavioral issue is evident in the supplied patch.

Finding count: none.

### Senior engineer — Passed

The stalled-response detection, session fencing, generation-bound retry authority, tool-activity fail-closed behavior, and parser metadata changes are coherently integrated and covered by focused regression tests. No concrete blocking correctness, security, concurrency, or maintainability defect is evident in the supplied patch.

Finding count: none.

The recovery path is substantially covered at the state and queue-admission levels, but the newly introduced Dispatcher-owned recovery scheduler is not exercised by production-path tests. Its durable source validation, permit issuance, and enqueue behavior are safety-critical and currently only indirectly assumed through a mocked scheduler.

backend/services/dispatcher.py
```diff
@@ -1344,12 +1350,104 @@ def revoke_context_retry_permit(
         ):
```
youchengsong, line 1350: **[MEDIUM] No integration coverage for the recovery scheduler**

Evidence: The patch adds `Dispatcher.enqueue_no_progress_recovery`, which performs the task/source/generation checks, derives the replay prompt, issues the volatile permit, and enqueues the internal retry. The existing instance-manager test was changed to replace this behavior with an `AsyncMock` scheduler, while the dispatcher tests cover generic queue admission but do not invoke this new method.

Impact: A regression in the scheduler's authorization checks, prompt selection, permit lifetime, or retry arguments could allow an unsafe replay or silently prevent recovery while all added tests still pass. This is the core safety boundary for the claimed automatic recovery.

Required fix: Add focused Dispatcher-level integration tests that call `enqueue_no_progress_recovery` against the test database and verify canonical source acceptance, rejection of stale/malformed/mismatched tasks, fresh-session retry arguments, exactly-once permit handling, and permit revocation when enqueue is rejected or raises.

Verification: Exercise the scheduler with a valid failed Claude task and source log, then assert the queued `no_progress_retry` message and permit; add negative cases for wrong generation/source shape, existing session, second attempt, and enqueue failure.

Finding fingerprint: `903c2417a526a2c6639a75fe62769ca70a783382a13002eb1bd5c885541e399b`

youchengsong (commented):
CCM automated review passed. The exact reviewed head `c6437d3030ede0d88190824e33ee984aa3de0ef5` is ready to merge.

PR #139, opened 2026-08-25
feat: make light theme the default

LianningLBY:
## Summary
- make `light` the default theme for first-time and invalid theme preferences
- keep existing `cc_theme` selections unchanged
- update initial HTML/PWA/iOS chrome colors to avoid a dark flash
- update theme regression coverage and project theme documentation

## Validation
- `tsc -b` passed
- focused theme/AppShell tests: 40 passed
- Vite production build passed
- full frontend suite: 907 passed, 7 skipped, 1 unrelated existing failure in `BrowserReviewPanel.test.tsx` (`本轮 PR99 前端验收` async polling expectation)

## Scope
- Branch is based on latest `origin/main`.
- No unrelated local worktree changes were included.


youchengsong (commented):

## Reviewer summaries

The default selection logic is coherent, but the static light bootstrap regresses the first paint for users who already selected dark.

The default theme fallback is implemented consistently, but browser chrome metadata is now permanently light even when an existing user preference selects the dark theme.

The persisted theme value is still honored after startup, but the new static light bootstrap causes users with an existing dark preference to visibly flash the wrong theme on every load.

frontend/index.html
```diff
@@ -1,12 +1,12 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-theme="light">
```
youchengsong, line 2: **[MEDIUM] Persisted dark users now flash light theme during startup**

Reviewer: `qa_engineer` · Category: `regression`

Evidence: The document is unconditionally initialized with `data-theme="light"`, while `frontend/src/config/theme.ts` later reads the existing `cc_theme` value and applies it. No pre-paint logic in the patch selects the stored theme before the initial render.

Impact: Users who previously selected `dark` see the light background/status chrome until the application JavaScript runs and switches back to dark, producing a visible theme flash and contradicting the promise to keep existing selections unchanged.

Required fix: Bootstrap the document theme from the persisted `cc_theme` before first paint (for example, a minimal inline script that validates the stored value and sets the matching data attribute/meta color), while retaining light as the fallback for first-time and invalid preferences.

Verification: Add an integration/browser startup test with `localStorage.cc_theme = 'dark'` that captures the initial document state before application hydration and verifies it starts dark; retain coverage for absent and invalid values starting light.

Finding fingerprint: `70a52a857db32a38c8cbc2dfe519fb0a7c0edd7cec0abf43d7f6c4175cb1d8d0`

youchengsong, line 2: **[MEDIUM] Existing dark-theme users now see a light-theme flash**

Reviewer: `principal_engineer` · Category: `regression`

Evidence: The document is unconditionally initialized with `data-theme="light"`, while `getTheme()` still honors an existing valid `cc_theme` such as `dark`; the stored preference is only applied after the frontend JavaScript runs.

Impact: Users with an existing dark selection, which the PR explicitly promises to keep unchanged, receive a light initial render and light browser chrome before the app switches back to dark. This is a visible theme regression on every cold load and can be especially noticeable on slower devices.

Required fix: Bootstrap the document theme before first paint by reading `cc_theme` and selecting it only when it is a valid supported theme, otherwise using `light`; update the initial theme-color metadata from the same bootstrap path (or otherwise provide an equivalent prepaint mechanism).

Verification: Add an integration/browser regression test with `localStorage.cc_theme = 'dark'` that loads the entry document and asserts the initial `data-theme` and theme-color are dark before application startup, plus a no-preference case asserting light.

Finding fingerprint: `5ebe8d25331758e1d096f24fa855ee8458266db543c64dc96c2f61de907227a8`

frontend/index.html
```diff
@@ -1,12 +1,12 @@
...
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover" />
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
-    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" />
-    <meta name="theme-color" content="#131316" />
+    <meta name="apple-mobile-web-app-status-bar-style" content="default" />
+    <meta name="theme-color" content="#e9e9ec" />
     <link rel="manifest" href="/manifest.json" />
```
youchengsong, line 11: **[MEDIUM] Static light browser chrome metadata ignores persisted theme**

Evidence: The patch hard-codes `theme-color` to `#e9e9ec`, changes the Apple status bar style to `default`, and sets the initial document theme to `light`. The supplied `theme.ts` changes do not update either metadata element when `applyTheme()` applies a stored theme, so a user with an existing `cc_theme=dark` preference receives light browser chrome and an initial light-theme paint.

Impact: Existing dark-theme users, whose selections the PR promises to preserve, can see a light flash on startup and retain a light status-bar/browser toolbar color while the app itself is dark; `default` status-bar content can also have insufficient contrast on dark backgrounds.

Required fix: Synchronize `meta[name="theme-color"]` and the Apple status-bar treatment with the active theme inside the theme application path, and apply the persisted theme before the first paint (for example with a minimal inline bootstrap) so the initial HTML state does not contradict stored preferences.

Verification: Add a test that stores `cc_theme=dark`, invokes the startup theme application, and asserts the document theme plus browser-chrome metadata/status-bar configuration are dark; cover the first-load bootstrap path as well.

Finding fingerprint: `047740e00e1dc1d003901bdce38b7e98278ceb9643bbc4a190e8dd0d9056d701`

frontend/index.html
LianningLBY: Fixed in 00530c8a. A synchronous pre-paint bootstrap now validates cc_theme and restores persisted dark themes before the application module runs, with light retained for missing or invalid values. Added startup regression coverage for persisted dark and missing or invalid preferences.

frontend/index.html
LianningLBY: Fixed in 00530c8a. Both the pre-paint bootstrap and runtime applyTheme path now synchronize theme-color and apple-mobile-web-app-status-bar-style with the active theme. Added tests for dark startup metadata and runtime dark/light switching.

frontend/index.html
LianningLBY: Fixed in 00530c8a. The entry document now applies a validated persisted theme, theme-color, and status-bar treatment synchronously before first paint. Existing dark users no longer receive the unconditional light bootstrap.

youchengsong (commented):

The panel found 5 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The bootstrap prevents the initial theme flash, but duplicates theme ownership and mishandles custom-theme chrome contrast during first paint.

The synchronous bootstrap improves persisted-theme startup, but it independently reimplements the theme registry and mishandles custom light themes.

Persisted built-in themes are covered, but the pre-paint bootstrap does not fully restore the supported custom theme.

frontend/index.html
```diff
@@ -1,12 +1,42 @@
...
+        const themes = {
+          dark: ['#131316', 'dark'],
+          light: ['#e9e9ec', 'light'],
+          feishu: ['#ecedef', 'light'],
+          apple: ['#f9f9f9', 'light'],
+          legacy: ['#030712', 'dark'],
+          ocean: ['#06131f', 'dark'],
+          forest: ['#07130d', 'dark'],
+          rose: ['#1a0b12', 'dark'],
+          custom: ['#131316', 'dark'],
```
youchengsong: **[MEDIUM] Custom themes always bootstrap with a dark status-bar scheme**

Evidence: The bootstrap declares `custom: ['#131316', 'dark']` and may replace only `themeColor` from `cc_theme_custom_bg`; `scheme` remains `dark` regardless of that color. In contrast, runtime `applyTheme` reads `document.documentElement.dataset.scheme` for custom themes, and its comment states that custom color scales are computed at runtime.

Impact: A persisted custom theme with a light background, such as `#ffffff`, initially receives `black-translucent` instead of the light-theme `default` status-bar mode. Because iOS may consume this metadata during launch, the later runtime correction can be too late and status-bar content can have incorrect contrast.

Required fix: Derive the custom scheme during the synchronous bootstrap using the same light/dark classification as the runtime custom-theme implementation, and set any required custom theme state before first paint.

Verification: Persist `cc_theme=custom` with a light custom background, run only the bootstrap, and assert the expected light scheme, `default` status-bar metadata, and restored custom-theme state; add a corresponding dark-background case.

Finding fingerprint: `2ca17b37e39dfc5afdc013b455993eb4e285e30959fe2b377e557f29498acb82`

frontend/index.html
```diff
@@ -1,12 +1,42 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover" />
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
-    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" />
-    <meta name="theme-color" content="#131316" />
+    <meta name="apple-mobile-web-app-status-bar-style" content="default" />
+    <meta name="theme-color" content="#e9e9ec" />
+    <script id="theme-bootstrap">
+      (() => {
+        const themes = {
```
youchengsong: **[MEDIUM] Bootstrap creates a second authoritative theme registry**

Reviewer: `senior_engineer` · Category: `architecture`

Evidence: The new inline `themes` object duplicates supported theme identifiers, chrome colors, schemes, and the `'light'` default independently of `THEME_OPTIONS`, `DEFAULT_THEME`, and runtime theme metadata in `theme.ts`. The changed project guidance still documents adding a theme through `theme.ts`, CSS, and optional icon registration, without this new registry, and the tests only exercise selected literals rather than checking registry parity.

Impact: Adding a theme or changing its chrome color/scheme through the documented extension path can leave the bootstrap stale. A valid persisted preference will then be treated as invalid or rendered with obsolete chrome until the application bundle runs, reintroducing the first-paint flash this patch is intended to eliminate.

Required fix: Generate or inject the synchronous bootstrap data from the authoritative theme configuration so identifiers, defaults, colors, and schemes have one change point. If a generated artifact is used, make generation/parity verification part of the build and tests.

Verification: Add a parity test proving that the bootstrap payload exactly matches every `THEME_OPTIONS` entry and `DEFAULT_THEME`, including color and scheme, then verify each persisted theme is restored correctly before application startup.

Finding fingerprint: `c8e67f6b912e856f76793747b3727ae1749d4b4045f4551d4bb6f41448fac40e`

frontend/index.html
```diff
@@ -1,12 +1,42 @@
...
+          dark: ['#131316', 'dark'],
+          light: ['#e9e9ec', 'light'],
+          feishu: ['#ecedef', 'light'],
```
youchengsong: **[MEDIUM] Theme registry is duplicated in the HTML bootstrap**

Reviewer: `principal_engineer` · Category: `architecture`

Evidence: The new inline bootstrap defines an independent `themes` object containing every theme and its chrome colors, while `frontend/src/config/theme.ts` remains the runtime authority for `THEME_OPTIONS`, schemes, and theme-color calculation.

Impact: Adding, renaming, or changing a theme requires synchronized edits in two unrelated modules. A missed update makes a persisted valid runtime theme fall back to light during bootstrap, causing incorrect first-paint chrome and a visible theme transition; the existing tests do not assert registry parity.

Required fix: Expose a small generated/static bootstrap mapping from the authoritative theme configuration, or add a parity mechanism that derives and verifies the HTML mapping from `THEME_OPTIONS` rather than maintaining a second hand-written registry.

Verification: Add a test that enumerates every `THEME_OPTIONS` value and verifies the bootstrap accepts it with the same scheme and default theme-color as `theme.ts`, including a newly added theme.

Finding fingerprint: `843ed75588a439a6d62d868eacc17f8c262684ab2611dd8ec22326ae0a03c4eb`

frontend/index.html
```diff
@@ -1,12 +1,42 @@
...
+          apple: ['#f9f9f9', 'light'],
```
youchengsong: **[MEDIUM] Custom theme status-bar scheme is hard-coded to dark**

Reviewer: `principal_engineer` · Category: `regression`

Evidence: The bootstrap maps `custom` to `['#131316', 'dark']` even when it reads an arbitrary valid `cc_theme_custom_bg` color. Runtime `applyTheme` explicitly derives the custom scheme from `document.documentElement.dataset.scheme`, indicating that custom backgrounds can have different contrast schemes.

Impact: Users with a light custom background get `black-translucent` Apple status-bar styling until the application JavaScript runs, producing incorrect first-paint contrast and potentially unreadable status-bar content.

Required fix: Use the same deterministic custom-background luminance/contrast calculation in the bootstrap (or persist the computed scheme alongside the custom color) and set the status-bar style from that result.

Verification: Bootstrap with `cc_theme=custom` and representative light and dark `cc_theme_custom_bg` values, then assert `apple-mobile-web-app-status-bar-style` is `default` for light backgrounds and `black-translucent` for dark backgrounds.

Finding fingerprint: `5c69723ba134ad0e809fe94092e013565eb2e09fd3209abf9abdf230359980d2`

frontend/index.html
youchengsong: **[MEDIUM] Pre-paint bootstrap only partially restores persisted custom themes**

Evidence: The bootstrap maps `custom` to a fixed dark scheme and restores only `cc_theme_custom_bg` for the meta color. It does not restore `data-scheme` or the custom theme's inline CSS variables. The production `applyTheme` code explicitly documents that custom color scales are calculated into runtime inline variables and reads `document.documentElement.dataset.scheme` when selecting the status-bar style. The added bootstrap tests cover dark, light, and invalid values but never custom.

Impact: A user with a persisted custom theme can receive the wrong custom palette and status-bar contrast during the first paint, until application initialization runs `applyTheme`. This contradicts the claimed restoration of existing theme selections before first paint and can produce the flash this patch intends to eliminate.

Required fix: Make the synchronous bootstrap restore the same persisted custom variables and scheme used by the runtime custom-theme path before styles paint, including deriving the status-bar style from that restored scheme.

Verification: Persist a custom theme configuration for both light and dark custom schemes, execute the bootstrap without calling runtime `applyTheme`, and assert `data-theme`, `data-scheme`, representative inline custom color variables, theme-color, and Apple status-bar style match the corresponding runtime result.

Finding fingerprint: `85c980fad91b4653fcc6a36f28008d1ebe13945e72e61791fae3effe6281b7dc`

frontend/index.html
LianningLBY: Fixed in 7f520d00. Theme metadata and DEFAULT_THEME now live in the single authoritative themeRegistry.ts module. Both runtime theme.ts and the generated pre-paint bundle import it. predev and prebuild regenerate the classic blocking script from the shared TypeScript entry, and tests enumerate every registry entry.

frontend/index.html
LianningLBY: Fixed in 7f520d00. The synchronous bootstrap now invokes the production custom-theme application path before first paint, restoring data-scheme, every generated custom CSS variable, theme-color, background/color-scheme state, and status-bar style. Tests compare all generated variables for both light and dark custom themes.

LianningLBY: Fixed in 7f520d00. The pre-paint path now calls the same applyCustomTheme implementation as runtime, so custom scheme classification, all managed CSS variables, theme color, background handling, and status-bar contrast are restored synchronously. Tests cover representative light and dark custom backgrounds.

frontend/index.html
LianningLBY: Fixed in 7f520d00. The handwritten HTML registry was removed. The committed bootstrap artifact is bundled from themeBootstrapEntry.ts, which imports themeRegistry.ts; predev and prebuild force regeneration. Parity tests enumerate all THEME_OPTIONS and verify scheme and chrome metadata.

frontend/index.html
LianningLBY: Fixed in 7f520d00. Custom bootstrap no longer hard-codes dark. It reuses applyCustomTheme, including the same OKLCh cutoff and data-scheme assignment used at runtime. Light and dark custom cases assert the expected Apple status-bar treatment.

youchengsong (commented):

## Reviewer summaries

The new prepaint path duplicates runtime theme application but omits the established custom-background capability, so persisted custom themes are not fully restored before paint.

The standard themes are restored consistently, but prepaint restoration is incomplete for custom themes using a background image.

### QA engineer — Passed

The patch implements the light default while preserving valid stored themes, restores themes before application startup, synchronizes browser chrome metadata, and adds focused coverage for defaults, invalid preferences, every registered theme, custom themes, and bootstrap generation.

Finding count: none.

frontend/src/config/themeBootstrap.ts
```diff
@@ -0,0 +1,37 @@
...
+  let theme = stored && THEME_MAP.has(stored as Theme) ? stored as Theme : DEFAULT_THEME;
+  let option = THEME_MAP.get(theme) ?? THEME_MAP.get(DEFAULT_THEME)!;
+  let themeColor: string = option.themeColor;
+  let scheme: 'dark' | 'light' = option.scheme;
+
+  if (theme === 'custom') {
+    try {
+      themeColor = applyCustomTheme();
+      scheme = document.documentElement.dataset.scheme === 'light' ? 'light' : 'dark';
+    } catch {
```
youchengsong, line 25: **[MEDIUM] Prepaint theme path omits persisted custom backgrounds**

Reviewer: `principal_engineer` · Category: `architecture`

Evidence: The runtime theme module imports both `applyCustomTheme` and `applyBgImage`, while the new bootstrap imports and calls only `applyCustomTheme`. The generated `public/theme-bootstrap.js` contains no application of the background URL or scrim; it only makes the root background transparent when the persisted custom-background flag is enabled.

Impact: Users whose persisted custom theme includes a background image initially receive an incomplete transparent/custom-color state, and the image appears only after the main application applies the second theme path. This reintroduces a first-paint flash and leaves two divergent implementations of theme restoration.

Required fix: Invoke the existing synchronous custom-background application capability from the prepaint path, or extract one shared browser-safe applicator used by both bootstrap and runtime theme application, then regenerate `public/theme-bootstrap.js`.

Verification: Persist a custom theme with background image enabled, execute the generated bootstrap before application startup, and assert that the root already has the expected background URL, scrim, and enabled-state attributes/styles; also verify switching to a non-custom theme clears them.

Finding fingerprint: `25c0e90840188b13abdd88af2d86be6bbc724647b719314181925e298991d527`

frontend/src/config/themeBootstrap.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  let theme = stored && THEME_MAP.has(stored as Theme) ? stored as Theme : DEFAULT_THEME;
+  let option = THEME_MAP.get(theme) ?? THEME_MAP.get(DEFAULT_THEME)!;
+  let themeColor: string = option.themeColor;
+  let scheme: 'dark' | 'light' = option.scheme;
+
+  if (theme === 'custom') {
+    try {
+      themeColor = applyCustomTheme();
```
youchengsong, line 22: **[MEDIUM] Prepaint bootstrap omits the persisted custom background image**

Evidence: The runtime theme module imports both `applyCustomTheme` and `applyBgImage`, while `applyPrepaintTheme` imports and calls only `applyCustomTheme`. The supplied generated bootstrap shows that `applyCustomTheme` detects `cc_theme_custom_has_bg` and makes the root background transparent, but it contains no corresponding application of the background URL/scrim. The new tests cover custom colors only and never enable a custom background.

Impact: A user whose existing selection is `custom` with a background image gets a transparent or incomplete initial canvas until the application runtime applies the image, producing the first-paint flash this change is intended to prevent.

Required fix: Apply the established custom-background capability during the custom prepaint path, or synchronously restore the same background URL and scrim properties there, with storage failures falling back consistently.

Verification: Configure a custom theme with its background-enabled flag, image URL, visibility/scrim settings, invoke `applyPrepaintTheme`, and assert that the root has the expected background properties immediately without running application startup.

Finding fingerprint: `4100b7bdc3bee3ce5af54f3d613b714ded58f3b9a7c39deebcb8706779c724dd`

frontend/src/config/themeBootstrap.ts
LianningLBY: Fixed in 25254961. The pre-paint path now restores a bounded localStorage snapshot of the last saved compressed image URL, CSS background URL, scrim, and data-has-bg state. IndexedDB remains authoritative and applyBgImage refreshes it after startup; missing snapshots safely fall back to the existing async path. Added regression coverage for the synchronous snapshot state.

frontend/src/config/themeBootstrap.ts
LianningLBY: Fixed in 25254961. Saving a background image now writes the same compressed data URL to a bootstrap snapshot while retaining IndexedDB as the source of truth. The classic blocking bootstrap applies that snapshot before first paint, and clearBgImage removes it. The generated public/theme-bootstrap.js was regenerated and tests/build pass.

youchengsong (commented):

## Reviewer summaries

The theme default and pre-paint restoration changes are cohesive, keep theme registration centralized, and add focused coverage for persisted themes, custom colors, browser chrome, and background snapshots.

Finding count: none.

The prepaint theme path is generally cohesive, but the background snapshot can become stale and imposes unbounded synchronous startup work.

The default-light behavior is covered, but prepaint background restoration is incompatible with existing custom-background data and can display stale or blank content.

frontend/src/config/customBg.ts
```diff
@@ -61,11 +62,13 @@ export async function loadBgImage(): Promise<string | null> {
 
 export async function saveBgImage(dataUrl: string): Promise<void> {
   await tx('readwrite', (s) => s.put(dataUrl, IMAGE_KEY));
+  try { localStorage.setItem(BOOTSTRAP_KEY, dataUrl); } catch { /* IDB remains the source of truth */ }
```
youchengsong: **[MEDIUM] Prepaint background cache is not synchronized with existing authoritative IDB data**

Reviewer: `qa_engineer` · Category: `regression`

Evidence: The new `ccm-theme-bootstrap-image` snapshot is written only by `saveBgImage()`, while the patch explicitly states that IndexedDB remains authoritative. Existing installations can therefore have `cc_theme_custom_has_bg=1` and a valid IDB image but no snapshot. In that state `applyCustomTheme()` makes the document background transparent because a background is flagged, while `applyBootstrapBgImage()` returns false, producing a blank backdrop until asynchronous IDB loading completes. If replacing an image makes `localStorage.setItem()` fail (for example due to quota), the caught error also leaves any previous snapshot intact, so bootstrap can show the old image before IDB supplies the new one. The added test seeds only the successful snapshot path and misses both states.

Impact: Users with backgrounds saved before this release can see a white/transparent first-paint flash, and users whose snapshot write fails can see a stale background flash. This contradicts the claimed first-paint restoration and breaks existing-data compatibility.

Required fix: Keep the bootstrap snapshot coherent with IDB: backfill or refresh it when the authoritative image is loaded, remove an old snapshot when a replacement write fails, and use the custom shell color rather than transparency when no valid bootstrap image was applied.

Verification: Seed the authoritative IDB image and `cc_theme_custom_has_bg=1` without the new localStorage key, then exercise bootstrap plus normal background loading and verify there is no transparent blank state and the snapshot is backfilled. Also seed an old snapshot, force `localStorage.setItem` to throw while replacing the IDB image, and verify bootstrap cannot apply the stale image.

Finding fingerprint: `f21bddfe55e9a064e78a20176c685a0eb423318f419b3ee304b335e92b0da326`

youchengsong: **[MEDIUM] Quota failure leaves a stale background snapshot active**

Evidence: saveBgImage() commits the new image to IndexedDB, then copies the complete data URL to localStorage inside a catch that does nothing. If setItem() fails—for example because a replacement image exceeds localStorage quota—the previous BOOTSTRAP_KEY remains while setHasBgImage(true) still runs. applyBootstrapBgImage() will consequently restore that old snapshot before applyBgImage() later reads the new authoritative IndexedDB value. The unrestricted full data URL is also synchronously read and parsed by the parser-blocking bootstrap.

Impact: Replacing a background can show the user's previous image on every subsequent first paint, and multi-megabyte snapshots can materially block page startup. Large but valid IndexedDB images commonly exceed localStorage quota after base64 expansion.

Required fix: Store a deliberately bounded bootstrap representation, such as a resized thumbnail. If updating that snapshot fails, remove the previous BOOTSTRAP_KEY so bootstrap cannot present stale content; keep IndexedDB authoritative.

Verification: Prepopulate BOOTSTRAP_KEY with an old image, make localStorage.setItem throw a quota error while saving a new image successfully to IndexedDB, and assert the old snapshot is removed and cannot be restored. Also verify the stored bootstrap representation respects the chosen size bound.

Finding fingerprint: `64ae99250b910c11052dadcf0460c97ac60b7445a3880430520cce89b737646e`

youchengsong (commented):
CCM automated review passed. The exact reviewed head `25f9753aa924f5df59879b6f615dc05248c24375` is ready to merge.

PR #140, opened 2026-08-25
fix(claude): recover oversized prompts safely

LianningLBY:
## Summary

- send large direct Claude prompts over stdin so Linux `MAX_ARG_STRLEN` does not fail with `E2BIG` before Claude Code starts
- recognize Claude Code's side-effect-free local `blocking_limit` envelope (`Prompt is too long`, zero API duration/usage) and reuse the fenced compact/retry recovery path
- lower the default proactive compaction threshold from 80% to 70% to leave room for the next message, tool output, and provider framing

## Safety

Claude automatic replay remains fail-closed. It requires the exact Task/Instance/retry/turn/source transport fence, a terminal `blocking_limit` result with `duration_api_ms=0` and zero usage, and no preceding tool/thinking/normal assistant activity.

## Tests

- `uv run pytest backend/tests/test_service_instance_manager.py -q -k 'large_prompt or large_claude_prompt or claude_prompt_too_long_preflight or codex_context_proof_loses_to_cancel'` — 4 passed
- `uv run pytest backend/tests/test_context_compaction.py backend/tests/test_claude_models.py backend/tests/test_api_settings_runtime.py -q` — 29 passed
- full `test_service_instance_manager.py`: 529 passed; 2 pre-existing environment-dependent failures caused by this host's `CLAUDE_BINARY=/home/ubuntu/.local/bin/claude` override and an existing PTY fixture expecting `/opt/claude-real`

## Reproduction evidence

A real 8010 test-environment Task with a ~660 KB initial prompt previously failed before Claude Code with `[Errno 7] Argument list too long`. Historical Wanghuan logs show the target Claude failure is a synthetic `invalid_request` followed by `terminal_reason=blocking_limit`, `duration_api_ms=0`, and zero token usage, proving it is a local preflight rejection with no model/tool side effects.

### DIFF
--- a/backend/services/instance_manager.py
+++ b/backend/services/instance_manager.py
@@ -14686,6 +14686,28 @@ class InstanceManager:
                 )
                 terminal_message = str(terminal_result.get("result") or "")
                 usage = terminal_result.get("usage")
+                required_usage_keys = (
+                    "input_tokens",
+                    "output_tokens",
+                )
+                optional_usage_keys = (
+                    "cache_creation_input_tokens",
+                    "cache_read_input_tokens",
+                )
+                usage_is_canonical_zero = (
+                    isinstance(usage, dict)
+                    and all(
+                        key in usage
+                        and type(usage[key]) is int
+                        and usage[key] == 0
+                        for key in required_usage_keys
+                    )
+                    and all(
+                        type(usage[key]) is int and usage[key] == 0
+                        for key in optional_usage_keys
+                        if key in usage
+                    )
+                )
                 if not (
                     terminal.event_type == "result"
                     and terminal.is_error is True
@@ -14693,17 +14715,9 @@ class InstanceManager:
                     and terminal_result.get("is_error") is True
                     and terminal_result.get("terminal_reason") == "blocking_limit"
                     and "prompt is too long" in terminal_message.lower()
+                    and type(terminal_result.get("duration_api_ms")) is int
                     and terminal_result.get("duration_api_ms") == 0
-                    and isinstance(usage, dict)
-                    and not any(
-                        int(usage.get(key) or 0)
-                        for key in (
-                            "input_tokens",
-                            "output_tokens",
-                            "cache_creation_input_tokens",
-                            "cache_read_input_tokens",
-                        )
-                    )
+                    and usage_is_canonical_zero
                 ):
                     return None
                 for row, raw in zip(rows[:-1], parsed_rows[:-1], strict=True):
--- a/backend/services/instance_manager.py
+++ b/backend/services/instance_manager.py
@@ -6124,6 +6124,8 @@ class InstanceManager:
                 "Codex isolated workflow execution forbids exec fallback"
             )
 
+        # PTY launches return above with the complete prompt. Only the direct
+        # exec transport reaches this argv-size fallback and owns stdin.
         claude_prompt_via_stdin = bool(
             provider == "claude"
             and len(prompt.encode("utf-8"))
@@ -14708,10 +14710,11 @@ class InstanceManager:
                     if row.event_type == "message":
                         if not (
                             row.role == "assistant"
-                            and row.is_error is False
+                            and row.is_error is True
                             and isinstance(raw, dict)
-                            and str(raw.get("error") or "").strip().lower()
-                            in {"", "invalid_request"}
+                            and raw.get("type") == "assistant"
+                            and raw.get("isApiErrorMessage") is True
+                            and raw.get("error") == "invalid_request"
                             and str(row.content or "").strip().lower()
                             == "prompt is too long"
                         ):
--- a/backend/services/instance_manager.py
+++ b/backend/services/instance_manager.py
@@ -14684,7 +14684,7 @@ class InstanceManager:
                 terminal_result = (
                     terminal_raw if isinstance(terminal_raw, dict) else {}
                 )
-                terminal_message = str(terminal_result.get("result") or "")
+                terminal_message = terminal_result.get("result")
                 usage = terminal_result.get("usage")
                 required_usage_keys = (
                     "input_tokens",
@@ -14696,6 +14696,9 @@ class InstanceManager:
                 )
                 usage_is_canonical_zero = (
                     isinstance(usage, dict)
+                    and set(usage).issubset(
+                        {*required_usage_keys, *optional_usage_keys}
+                    )
                     and all(
                         key in usage
                         and type(usage[key]) is int
@@ -14714,6 +14717,7 @@ class InstanceManager:
                     and terminal_result.get("type") == "result"
                     and terminal_result.get("is_error") is True
                     and terminal_result.get("terminal_reason") == "blocking_limit"
+                    and isinstance(terminal_message, str)
                     and "prompt is too long" in terminal_message.lower()
                     and type(terminal_result.get("duration_api_ms")) is int
                     and terminal_result.get("duration_api_ms") == 0
--- a/backend/services/instance_manager.py
+++ b/backend/services/instance_manager.py
@@ -72,6 +72,12 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+# Linux rejects a single argv entry around 128 KiB (MAX_ARG_STRLEN), before
+# Claude Code can apply its own context-window guard. Keep ample room for
+# encoding/platform variance and send only large direct-Claude prompts over
+# stdin; ordinary launches retain their established argv shape.
+_CLAUDE_PROMPT_STDIN_THRESHOLD_BYTES = 64 * 1024
+
 
 async def _fence_worker_runtime_mutation(
     db: AsyncSession,
@@ -6118,9 +6124,15 @@ class InstanceManager:
                 "Codex isolated workflow execution forbids exec fallback"
             )
 
+        claude_prompt_via_stdin = bool(
+            provider == "claude"
+            and len(prompt.encode("utf-8"))
+            >= _CLAUDE_PROMPT_STDIN_THRESHOLD_BYTES
+        )
         cmd = self._build_command(
             provider=provider,
             prompt=prompt,
+            claude_prompt_via_stdin=claude_prompt_via_stdin,
             model=model,
             resume_session_id=resume_session_id,
             effort_level=effort_level,
@@ -6279,6 +6291,8 @@ class InstanceManager:
                 "env": env,
                 "limit": 10 * 1024 * 1024,
             }
+            if claude_prompt_via_stdin:
+                spawn_kwargs["stdin"] = asyncio.subprocess.PIPE
             if os.name == "posix":
                 spawn_kwargs["start_new_session"] = True
             await admit_external_launch("claude_exec")
@@ -6300,6 +6314,12 @@ class InstanceManager:
                         else None
                     ),
                 )
+                if claude_prompt_via_stdin:
+                    await self._write_direct_claude_prompt(
+                        instance_id,
+                        process,
+                        prompt,
+                    )
             except BaseException:
                 if task_private_tmpdir is not None:
                     exact_process = self.processes.get(instance_id)
@@ -12381,15 +12401,18 @@ class InstanceManager:
         claude_unrestricted_settings_path: Path | None = None,
         claude_unrestricted_tools: Sequence[str] | None = None,
         claude_unrestricted_allowed_rules: Sequence[str] | None = None,
+        claude_prompt_via_stdin: bool = False,
     ) -> list[str]:
         """Build the subprocess command for a supported coding-agent CLI."""
         if provider == "claude":
             cmd = [
                 settings.claude_binary,
-                "-p", prompt,
+                "-p",
                 "--output-format", "stream-json",
                 "--verbose",
             ]
+            if not claude_prompt_via_stdin:
+                cmd.insert(2, prompt)
             if claude_isolation_settings_path is not None:
                 from backend.services.task_agent_isolation import (
                     CLAUDE_TASK_BUILTIN_TOOLS,
@@ -14526,20 +14549,24 @@ class InstanceManager:
         expected_turn_generation: int | None = None,
         expected_started_at: datetime | None = None,
     ) -> _ContextPreflightPermit | None:
-        """Prove a direct Codex chat overflow happened before agent activity.
+        """Prove a direct chat overflow happened before agent activity.
 
         This is deliberately independent of stderr and rendered message text.
         Only the current exact source, its committed runtime transport, and a
-        strict durable ``turn.failed`` envelope can authorize compaction and
-        replay.  Missing, stale, malformed, or mixed-turn evidence fails
-        closed.
+        strict durable provider envelope can authorize compaction and replay.
+        Missing, stale, malformed, or mixed-turn evidence fails closed.
         """
 
+        provider = (
+            str(params.get("provider") or "").strip().lower()
+            if isinstance(params, dict)
+            else ""
+        )
         if (
             type(task_id) is not int
             or task_id <= 0
             or not isinstance(params, dict)
-            or str(params.get("provider") or "").strip().lower() != "codex"
+            or provider not in {"codex", "claude"}
         ):
             return None
         requested_source_id = params.get("source_log_id")
@@ -14571,7 +14598,7 @@ class InstanceManager:
             if (
                 task is None
                 or instance is None
-                or (task.provider or "claude").lower() != "codex"
+                or (task.provider or "claude").lower() != provider
                 or task.status not in {"executing", "in_progress"}
                 or task.instance_id != instance_id
                 or task.retry_count != expected_retry_count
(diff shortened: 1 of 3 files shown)


youchengsong (commented):

## Reviewer summaries

The Claude replay authorization gate does not fully enforce the claimed absence of prior assistant activity.

Large prompts are removed from every Claude command, but the new stdin delivery is wired only to the exec transport, leaving the supported PTY transport without the prompt.

The large-prompt stdin path is covered, but the safety-critical Claude recovery claim is only tested at the permit helper and not through compaction and replay.

backend/tests/test_service_instance_manager.py
```diff
@@ -22160,6 +22207,102 @@ async def readline():
     assert "already wrapped history" not in retry["prompt"]
```
youchengsong, line 22209: **[MEDIUM] Claude overflow recovery is not tested end to end**

Evidence: The sole Claude blocking-limit test directly invokes `_chat_structured_context_preflight_rejection` and only asserts that it returns a permit. It never exercises the production failure handler, context compaction, session replacement, or retry dispatch claimed by the PR. It also provides no integration-level proof that disallowed preceding tool, thinking, or normal assistant activity prevents automatic replay.

Impact: The helper can accept the new envelope while Claude failures still fail to enter the recovery path, compact incorrectly, or replay after side effects. Those regressions would pass the added test despite violating both the primary behavior and fail-closed safety claim.

Required fix: Add focused tests through the production failure/recovery entry point: one valid Claude `blocking_limit` envelope must compact and schedule exactly one fenced retry with the expected session transition, while an otherwise identical envelope containing prior agent activity must not compact or replay.

Verification: Drive the actual failure handler with persisted task, source, and foreground logs; assert the valid case performs compaction and one retry, and parameterize tool, thinking, and normal assistant predecessor events to assert zero retries and unchanged task/session state.

Finding fingerprint: `60f842741ecd687775689c07850aa6a8b70fb5bbdae11557d604d1feed96d5d0`

backend/services/instance_manager.py
```diff
@@ -6118,9 +6124,15 @@ async def admit_claude_pty_transport() -> None:
                 "Codex isolated workflow execution forbids exec fallback"
```
youchengsong, line 6124: **[MEDIUM] Large Claude prompts are dropped on the PTY transport**

Evidence: `claude_prompt_via_stdin` is derived solely from provider and prompt size and is passed to `_build_command`, which consequently omits the prompt from argv. However, `stdin=PIPE` and `_write_direct_claude_prompt(...)` are added only in the `asyncio.create_subprocess_exec` path. The same changed code explicitly recognizes `claude_pty` as a supported source transport, and the complete patch adds no corresponding delivery through that PTY path. The new test mocks only `create_subprocess_exec`, so it does not exercise this case.

Impact: A Claude launch selecting `claude_pty` with a prompt at or above 64 KiB starts with `-p` but receives neither an argv prompt nor the new stdin write. It can wait for input, fail, or execute without the requested task, so the oversized-prompt fix regresses a supported transport.

Required fix: Select stdin prompt mode only after the actual transport is known and restrict it to `claude_exec`, or implement equivalent prompt delivery and EOF handling for `claude_pty`.

Verification: Force a Claude PTY launch with a prompt at the threshold, assert the complete UTF-8 prompt is delivered through the PTY input and EOF is signaled, and assert the process does not start with a promptless command unless that delivery path is active.

Finding fingerprint: `93305ea98ef4e419afa3646be95ba728eb7f9824be46612c3bd5cfc75e097799`

backend/services/instance_manager.py
```diff
@@ -14647,6 +14678,59 @@ async def _chat_structured_context_preflight_rejection(
...
+                            "cache_creation_input_tokens",
+                            "cache_read_input_tokens",
+                        )
+                    )
+                ):
+                    return None
+                for row, raw in zip(rows[:-1], parsed_rows[:-1], strict=True):
+                    if row.event_type == "message":
+                        if not (
+                            row.role == "assistant"
```
youchengsong, line 14730: **[MEDIUM] Replay gate accepts an ordinary assistant message as preflight evidence**

Reviewer: `principal_engineer` · Category: `correctness`

Evidence: For a preceding message, the new Claude branch accepts `raw.get("error")` in `{ "", "invalid_request" }`. A normal assistant row with no error marker, `is_error=False`, and content `Prompt is too long` therefore satisfies the gate. This conflicts with the stated requirement that replay have no preceding normal assistant activity. The added test covers only the `invalid_request` success case and does not exercise this permissive path.

Impact: Compaction and automatic replay can be authorized after assistant activity rather than only after Claude's synthetic local rejection. That weakens the fail-closed recovery boundary and can cause an already-started turn to be replayed.

Required fix: Require the preceding synthetic message to carry the exact `invalid_request` marker, along with any other stable fields needed to distinguish it from a normal assistant message; otherwise return `None`.

Verification: Add a negative test using the otherwise-valid Claude `blocking_limit` envelope but a preceding assistant message with a missing or empty `error` field, and assert that `_chat_structured_context_preflight_rejection` returns `None`.

Finding fingerprint: `4cb7600136d7c466c00b019423adb3890a32755550dd6c5052f0597528d74ebb`

backend/tests/test_service_instance_manager.py
LianningLBY: Addressed in 45f8234c. The Claude test now drives the real `_consume_output` failure handler with actual stream-json assistant/result envelopes, then asserts one compaction, one fenced `compact_retry`, the expected summary/current-message payload, and session clearing. A parameterized production-path test covers preceding tool use, thinking, normal assistant output, and a missing `invalid_request` marker; every case asserts zero compactions/retries and preserves the original session.

backend/services/instance_manager.py
LianningLBY: The reported prompt-drop path is not reachable: the Claude PTY branch returns from `launch()` before `claude_prompt_via_stdin` is computed or `_build_command` is called, and passes the complete prompt directly to `_launch_pty`. Only the direct `claude_exec` branch reaches stdin mode. Commit 45f8234c adds an explicit control-flow comment and forces a >64 KiB PTY launch in the test, asserting the exact full prompt reaches `_launch_pty`.

backend/services/instance_manager.py
LianningLBY: Addressed in 45f8234c. Replay now requires the exact synthetic envelope and persisted shape: `type=assistant`, `isApiErrorMessage=true`, `error=invalid_request`, and `row.is_error=true`. This also fixes the real parser path, because `StreamParser` persists an API error message with `is_error=true`; the previous helper-only fixture incorrectly used false. The production failure-handler negative test verifies an otherwise matching prompt-too-long envelope without the error marker does not compact or replay.

youchengsong (commented):

## Reviewer summaries

The stdin transport fits the existing launch path, but the automatic replay authorization does not fail closed for malformed usage envelopes.

Claude replay fencing is improved, but usage validation does not fail closed for incomplete or malformed evidence.

The oversized-prompt transport is covered, but malformed Claude usage evidence can raise inside the safety-critical replay proof instead of failing closed.

backend/services/instance_manager.py
```diff
@@ -14647,6 +14680,60 @@ async def _chat_structured_context_preflight_rejection(
...
+                )
+                terminal_message = str(terminal_result.get("result") or "")
+                usage = terminal_result.get("usage")
+                if not (
+                    terminal.event_type == "result"
+                    and terminal.is_error is True
+                    and terminal_result.get("type") == "result"
+                    and terminal_result.get("is_error") is True
+                    and terminal_result.get("terminal_reason") == "blocking_limit"
+                    and "prompt is too long" in terminal_message.lower()
```
youchengsong, line 14721: **[MEDIUM] Replay proof accepts incomplete usage and can raise on malformed counters**

Reviewer: `principal_engineer` · Category: `correctness`

Evidence: The Claude permit treats every absent counter as zero via `int(usage.get(key) or 0)` and accepts any `usage` dictionary, including `{}`. Non-numeric values such as `"unknown"` instead raise `ValueError`, despite the method's contract that malformed evidence returns no permit. Python coercion also accepts non-canonical representations such as `"0"`, while `duration_api_ms == 0` accepts `False`.

Impact: The safety-sensitive compact/replay path can authorize a retry without affirmative zero-usage evidence, or abort terminal processing instead of rejecting a malformed provider envelope. This weakens the stated fail-closed recovery and idempotency guarantee.

Required fix: Validate the Claude terminal envelope without coercion: require canonical numeric zero values for the mandatory input/output counters and `duration_api_ms`, reject booleans and malformed values, validate any optional cache counters when present, and return `None` for every schema mismatch.

Verification: Add direct or end-to-end cases proving no compaction or enqueue occurs for an empty usage object, missing mandatory counters, boolean/string counters or duration, and non-numeric counter values; retain a passing case with canonical integer zeros and optional cache counters omitted or explicitly zero.

Finding fingerprint: `e19fa11750df1c6b7c1adc306418c2c6be433f5c82db03554e0c03adea54ba82`

backend/services/instance_manager.py
```diff
@@ -14647,6 +14680,60 @@ async def _chat_structured_context_preflight_rejection(
...
+                    and terminal_result.get("duration_api_ms") == 0
+                    and isinstance(usage, dict)
+                    and not any(
```
youchengsong: **[MEDIUM] Usage validation can accept missing evidence or raise on malformed values**

Evidence: The replay permit requires only `isinstance(usage, dict)` and evaluates `int(usage.get(key) or 0)`. Missing token fields and falsey nonnumeric values are treated as proven zero, so even an empty usage object passes. Truthy malformed values such as `{"input_tokens":"unknown"}` raise `ValueError` instead of returning `None`, contrary to the function's stated fail-closed handling of malformed evidence.

Impact: An incomplete provider envelope can authorize automatic compaction and replay without durable proof of zero model usage, while malformed output can escape the proof routine and disrupt terminal processing instead of being rejected safely.

Required fix: Validate the required usage schema explicitly: require each authoritative token field to be present with a non-boolean integer value equal to zero, or use a provider-schema-aware validator if Claude legitimately omits defined zero fields. Any missing, invalid, or nonzero value must return `None` without raising.

Verification: Add parameterized proof tests covering an empty usage object, each missing required field, `None`, booleans, numeric strings, arbitrary strings, collections, and nonzero values. Assert every invalid envelope returns no permit and that the output consumer neither compacts nor requeues.

Finding fingerprint: `488b1f64661b3a539d79d4d9c85f1d42f4d64e608dd668926c86c6e8b026222a`

backend/services/instance_manager.py
```diff
@@ -14647,6 +14680,60 @@ async def _chat_structured_context_preflight_rejection(
...
+                        int(usage.get(key) or 0)
```
youchengsong: **[MEDIUM] Malformed usage values can crash the replay proof**

Evidence: The Claude proof converts durable provider-controlled JSON with `int(usage.get(key) or 0)`. Non-numeric values such as `"unknown"`, objects, or lists raise `ValueError` or `TypeError`; fractional values such as `0.5` are truncated to zero and can incorrectly satisfy the zero-usage condition. This contradicts the function's stated rule that malformed evidence fails closed.

Impact: A malformed or version-changed Claude result can interrupt output-consumer error handling instead of returning no permit, and some nonzero malformed usage can authorize automatic compaction and replay without exact proof of zero API usage.

Required fix: Validate each present usage field without coercion: require a non-boolean integer value equal to zero, and return `None` for every other type or value.

Verification: Feed the production recovery path terminal results containing usage values `"unknown"`, `{}`, `0.5`, and `True`; verify consumption completes, compaction and enqueue are not called, and the existing session remains unchanged.

Finding fingerprint: `6232f9ab972d5a1097933cec112f6176af0d64e4ad6f1e837fcd211e1ec69961`

backend/services/instance_manager.py
LianningLBY: Fixed in 8bc420eb. Claude replay proof now requires canonical usage evidence: input_tokens and output_tokens must both be present native integers equal to 0; optional cache counters, when present, must also be native integer 0. duration_api_ms must be a native integer 0. Missing fields, booleans, strings, floats, malformed objects, and nonzero values return None without raising.

backend/services/instance_manager.py
LianningLBY: Fixed in 8bc420eb. Removed all int() coercion from the Claude replay proof and added production failure-handler coverage for empty/missing/None/bool/string/fractional/arbitrary/nonzero token values. Each case asserts no compaction and no enqueue.

backend/services/instance_manager.py
LianningLBY: Fixed in 8bc420eb. Malformed usage and duration values now fail closed instead of raising. The recovery-path tests cover invalid cache values, boolean duration, and string duration in addition to malformed token counters.

youchengsong (commented):

## Reviewer summaries

The transport and replay changes fit the existing manager flow, but the fail-closed Claude replay proof still accepts malformed terminal envelopes.

### Senior engineer — Passed

The prompt transport, strict Claude preflight proof, fail-closed validation, and focused regression tests are coherent and adequately scoped. No concrete correctness, security, concurrency, or maintainability defect is evidenced by the supplied patch.

Finding count: none.

The transport path is meaningfully tested, but the safety-critical replay proof still accepts non-canonical usage envelopes that do not prove zero usage.

backend/services/instance_manager.py
```diff
@@ -14647,6 +14680,74 @@ async def _chat_structured_context_preflight_rejection(
 
             terminal = rows[-1]
             terminal_raw = parsed_rows[-1]
+            if provider == "claude":
+                terminal_result = (
+                    terminal_raw if isinstance(terminal_raw, dict) else {}
+                )
```
youchengsong, line 14686: **[HIGH] Claude preflight proof accepts non-string terminal results**

Reviewer: `principal_engineer` · Category: `security`

Evidence: The proof assigns `terminal_message = str(terminal_result.get("result") or "")` and then authorizes replay when the converted value contains `"prompt is too long"`. There is no `isinstance(..., str)` check, so a result such as `{ "result": { "detail": "Prompt is too long" }, ... }` satisfies the phrase test despite not being the canonical Claude envelope.

Impact: A malformed or unexpected Claude output can trigger automatic compaction and replay even though the required local preflight error was not proven. This violates the patch's fail-closed safety boundary and can replay a turn without reliable evidence that no model/tool activity occurred.

Required fix: Require `result` to be an actual string before matching it (and preferably validate the canonical message/result shape), returning `None` for non-string values.

Verification: Add a regression case with `result` as a dict/list (containing the phrase) and assert that neither compaction nor enqueue/replay occurs.

Finding fingerprint: `89c949f1e3abcc6cae2f0d39e047772121860d61bb357eea4cf4a902f18d8012`

backend/services/instance_manager.py
```diff
@@ -14647,6 +14680,74 @@ async def _chat_structured_context_preflight_rejection(
...
+                        type(usage[key]) is int and usage[key] == 0
+                        for key in optional_usage_keys
+                        if key in usage
+                    )
+                )
+                if not (
+                    terminal.event_type == "result"
+                    and terminal.is_error is True
+                    and terminal_result.get("type") == "result"
+                    and terminal_result.get("is_error") is True
```
youchengsong, line 14718: **[MEDIUM] Replay proof ignores unknown nonzero usage counters**

Evidence: `usage_is_canonical_zero` requires two known counters to be zero and checks two optional counters only when present, but accepts every other key without validation. For example, `{"input_tokens":0,"output_tokens":0,"future_input_tokens":1}` passes. The added negative tests cover missing, mistyped, and nonzero known fields, but not unknown nonzero fields.

Impact: If Claude adds another usage counter, or a malformed stored envelope contains one, the code may authorize compaction and automatic replay without proving the claimed zero-usage/no-side-effect condition, risking duplicate work or side effects.

Required fix: Make the usage-envelope check fail closed: validate against an explicit supported schema and reject unknown fields, or explicitly validate every additional documented usage field before permitting replay.

Verification: Add an end-to-end negative replay case with required counters at zero plus an unknown nonzero usage field; assert that neither `_compact_session` nor `enqueue_message` is called and the existing session remains unchanged.

Finding fingerprint: `7f35756c8123eb2bd6617cf3740db3cb7a185cc157851729034887f41a017f35`

LianningLBY: Fixed in 8b6d33cb. The usage validator now requires an explicit supported key set and rejects unknown fields before replay authorization. Added an end-to-end case with `future_input_tokens=1`; compaction and enqueue remain untouched.

backend/services/instance_manager.py
LianningLBY: Fixed in 8b6d33cb. The proof now keeps the terminal result uncoerced, requires `isinstance(result, str)`, and only then performs the Prompt is too long match. Added an end-to-end recovery negative case with a dict result containing the phrase; it performs no compaction or enqueue.

youchengsong (commented):
CCM automated review passed. The exact reviewed head `8b6d33cbdbfdf0d7878113a9a3212bc768abde65` is ready to merge.

PR #143, opened 2026-08-29
fix(browser-review): launch bundled Chromium safely

Citrus086:
## Problem

Frontend review defaults to Playwright's bundled Chromium, but the runtime passes `browser_channel="chromium"` as Playwright's `channel` option. Recent Playwright versions then resolve the optional headless-shell path, and review startup fails when that browser is not installed.

On Linux hosts with unprivileged user namespaces disabled, Chromium also needs the matching setuid `chrome_sandbox` helper exposed through `CHROME_DEVEL_SANDBOX`. Without it, launch fails even after installing Playwright browsers.

## Fix

- Treat `chromium` as the bundled Playwright browser and only pass `channel="chrome"` for explicit system Chrome selection.
- Automatically configure the matching Playwright `chrome_sandbox` helper when it exists and has the setuid bit.
- Add regression coverage for the browser-channel mapping.

The implementation keeps Chromium sandboxing enabled and does not fall back to `--no-sandbox`.

## Verification

- `pytest -q backend/tests/test_browser_review.py backend/tests/test_browser_review_defaults.py backend/tests/test_browser_review_jobs.py`
- 56 passed
- Real Playwright Chromium launch verified with the sandbox helper configured.


youchengsong (commented):

## Reviewer summaries

The channel mapping is covered, but the safety-critical sandbox configuration path is untested and lacks proof that the launch environment is configured correctly across the supported runtime conditions.

### Senior engineer — Passed

The browser-channel mapping and sandbox-helper configuration are coherent with the stated runtime failure, preserve sandboxing, and include focused regression coverage. No concrete blocking issue is evident in the supplied patch.

Finding count: none.

The bundled-browser channel mapping is covered, but the new sandbox setup can fail before Playwright launches when no environment mapping is present.

Reviewed commit: `4234021a390ea467e9ff3446e16cf8d16ad2b7b3`.

backend/services/browser_review.py
```diff
@@ -780,6 +780,21 @@ async def _browser_page(
                 options,
                 proxy_url=proxy.url if proxy is not None else None,
             )
+            if options.browser_channel != "chrome":
+                # Playwright's headless-shell package does not ship its own
+                # setuid sandbox. On Linux hosts where unprivileged user
```
youchengsong: **[HIGH] Sandbox helper configuration has no regression coverage**

Reviewer: `principal_engineer` · Category: `testing`

Evidence: The patch adds conditional discovery of `chrome_sandbox`, checks its setuid bit, and mutates `launch_options["env"]["CHROME_DEVEL_SANDBOX"]`, but the added test only verifies the `browser_channel` mapping. No test exercises the helper-present/setuid case, helper-missing case, or the resulting arguments passed to `playwright.chromium.launch`.

Impact: The change is intended to make Chromium launch safely on hosts with disabled unprivileged user namespaces. A path derivation, permission check, or environment-shape regression could either prevent launches on those hosts or leave the claimed sandbox configuration unverified; the current suite would not detect it.

Required fix: Add focused tests around `_browser_page` or the launch setup that mock `playwright.chromium.executable_path` and filesystem metadata, asserting that a setuid sibling helper is exported as `CHROME_DEVEL_SANDBOX`, while absent or non-setuid helpers do not alter the environment and no `--no-sandbox` fallback is introduced.

Verification: Run the browser-review tests with mocked helper-present, helper-absent, and non-setuid scenarios, and assert the exact `launch` environment for each case.

Finding fingerprint: `9ca6ac43d88678ac51a755b1fbe2bf171984f4921ccaf5902f71ab38b9ebb178`

backend/services/browser_review.py
```diff
@@ -780,6 +780,21 @@ async def _browser_page(
...
+                # namespaces are disabled, point Chromium at the matching
+                # Playwright sandbox helper instead of weakening the browser
+                # with --no-sandbox.
+                sandbox = Path(playwright.chromium.executable_path).with_name(
+                    "chrome_sandbox"
```
youchengsong: **[HIGH] Sandbox setup assumes launch_options contains an env mapping**

Evidence: The new branch unconditionally assigns `launch_options["env"]["CHROME_DEVEL_SANDBOX"]`, but this code does not initialize or default the `env` key before indexing it. The launch-options builder shown in the patch only adds launch arguments and the optional channel.

Impact: On a normal bundled Chromium launch where no explicit `env` mapping was previously supplied, finding a valid `chrome_sandbox` helper raises `KeyError` and prevents every browser review from starting, directly defeating the fix.

Required fix: Create or reuse an environment mapping before assignment, preserving the inherited process environment (for example, `env = launch_options.setdefault("env", os.environ.copy())` and then set `env["CHROME_DEVEL_SANDBOX"]`).

Verification: Add a regression test that builds launch options without an `env` key, mocks an executable-adjacent setuid `chrome_sandbox`, runs the launch-option preparation path, and asserts launch proceeds with the helper path in the resulting environment.

Finding fingerprint: `3b2e0a2228a7402f1787d79e7bac6a4f3f5998bc0d15a4dd9866e4cf290b57ac`

Citrus086:
Addressed the blocking sandbox findings in commit d0a618df. Extracted bundled Chromium sandbox setup into a testable helper and added coverage for setuid helper present, helper absent, and non-setuid helper cases. Tests also assert the exact launch environment and that --no-sandbox is never added. Verification: 59 passed.

youchengsong (commented):
CCM automated review passed. The exact reviewed head `d0a618df920fcba657d3eb6ab331b682bc0b3815` is ready to merge.

youchengsong (commented):
CCM automated review passed. The exact reviewed head `e5b427624d786871461f86ea9869c953c6b0c151` is ready to merge.

PR #149, opened 2026-09-16
fix: prevent managed Claude tasks from creating orphan cron turns

Citrus086:
## Summary

- disable Claude-native CronCreate/CronDelete/CronList/ScheduleWakeup for managed CCM Tasks
- keep scheduling through CCM Monitor/Delivery durable paths
- add regression coverage for the generated disallowed-tools list

## Why

A foreground PTY/API disconnect could terminalize a Task while a Claude-native recurring cron remained alive in the resumed session. The cron then generated autonomous turns and competed with later user messages. Managed Task lifecycles must not contain provider-local autonomous producers.

## Tests

- `pytest -q backend/tests/test_service_instance_manager.py -k 'build_command_claude_task_disables_interactive_plan_mode or build_command_claude_enable_workflows_true or build_command_claude_enable_workflows_default'`\n- `pytest -q backend/tests/test_task_agent_isolation.py -k 'permission or builtin or tool'`


youchengsong (commented):

## Reviewer summaries

The change is narrowly scoped to managed Claude task isolation, uses the existing disallowed-tools mechanism, and adds regression coverage for the generated list. No concrete architectural or behavioral issue is evidenced by the supplied patch.

Finding count: none.

### Senior engineer — Passed

The managed-task disallowed-tools list is extended consistently to cover all provider-local cron and wakeup capabilities, with regression coverage asserting the generated set. No concrete correctness, security, or maintainability issue is evident in the supplied patch.

Finding count: none.

The new command restriction prevents future managed tasks from invoking Claude cron tools, but does not address cron jobs already persisted in resumable sessions.

Reviewed commit: `e34e17b46b1b2566adde32c85b6ac4d11aa8ee7a`.

backend/services/task_agent_isolation.py
```diff
@@ -51,6 +51,16 @@
 CLAUDE_TASK_INTERACTIVE_DISALLOWED_TOOLS = (
     "EnterPlanMode",
     "ExitPlanMode",
+    # Native Claude cron/wakeup turns run inside the provider session rather
```
youchengsong, line 54: **[MEDIUM] Existing provider cron jobs are not cleaned up or neutralized**

Reviewer: `qa_engineer` · Category: `regression`

Evidence: The patch only adds CronCreate/CronDelete/CronList/ScheduleWakeup to the generated --disallowedTools list. There is no cleanup, migration, or resume-time handling for cron jobs that were created before this change; the PR description explicitly identifies such jobs as surviving in resumed sessions.

Impact: A managed task/session created before deployment can still contain a provider-local recurring producer, so it may continue generating autonomous turns after terminalization and compete with later user messages despite this fix.

Required fix: On managed-task resume/startup, detect and remove or otherwise disable existing Claude-native cron schedules (or explicitly invalidate affected sessions) in addition to blocking creation of new ones.

Verification: Create a managed Claude session with a pre-existing native cron, terminalize/disconnect it, resume or restart it through the normal lifecycle, and assert the cron cannot generate another turn; also verify normal CCM Monitor/Delivery scheduling remains functional.

Finding fingerprint: `0a6193bb9a126e145bc02279b0b0035a286802d14a903e6ef071746e60ecee47`

Citrus086:
Addressed the blocking QA finding in commit `a23bd8c7`.

Managed Claude launches now set `CLAUDE_CODE_DISABLE_CRON=1` for both direct subprocess and PTY paths, in addition to denying `CronCreate/CronDelete/CronList/ScheduleWakeup`. This disables provider-local schedules already persisted in resumable sessions, including on resume. Added a regression test for the managed launch environment.

Targeted tests pass:
- 2 InstanceManager cron/managed-task tests
- 3 task isolation permission tests

youchengsong (commented):

## Reviewer summaries

The managed-launch path is covered, but the resume/config path that is central to preventing persisted cron schedules from returning has no regression proof.

### Senior engineer — Passed

The patch consistently blocks creation of new provider-local schedules and disables persisted schedules for managed Claude tasks across the supplied launch paths. No concrete correctness, security, or maintainability defect is evident in the injected material.

Finding count: none.

The deny-list is covered, but the safety-critical resume protection is not behaviorally verified; the only new test checks an environment assignment during launch.

Reviewed commit: `a23bd8c78806b2fd8b65c9959b4da27efd3eea6b`.

backend/services/instance_manager.py
```diff
@@ -6582,6 +6582,13 @@ async def admit_claude_pty_transport() -> None:
 
         # Disable CC's auto-compact — CCM manages context/compaction itself
         env["DISABLE_AUTO_COMPACT"] = "true"
+        # Managed Tasks must not revive provider-local cron schedules from a
+        # resumable Claude session. CCM scheduling is handled by Dispatcher and
+        # durable Monitor/Delivery records instead.
+        if provider == "claude" and task_id is not None:
```
youchengsong: **[MEDIUM] Resume-time cron suppression is untested**

Evidence: The patch claims CLAUDE_DISABLE_CRON prevents persisted native schedules from being revived and adds a separate override in _patched_build_config, but test_managed_claude_launch_disables_native_cron only asserts mock_exec.call_args.kwargs["env"][CLAUDE_DISABLE_CRON] == "1". No test exercises the patched config/resume path or verifies that an existing schedule is actually suppressed.

Impact: A regression in the resume/config path, or an ineffective provider switch, could leave pre-existing Claude cron producers active after a managed task resumes—the exact orphan-turn failure this PR is intended to prevent—while the suite remains green.

Required fix: Add a focused test covering resumable-session/_patched_build_config construction (including task_id) and asserting the disable-cron override reaches the spawned Claude process; where feasible, verify a persisted schedule is not reactivated.

Verification: Unit-test the resume path with a mocked Claude config/process and assert CLAUDE_DISABLE_CRON=1 in the final environment/overrides, in addition to the existing initial-launch test.

Finding fingerprint: `3c624f81204c07cc6d8daf051f833ab72fd57700abaa061ac338cf5e9cb57802`

backend/services/instance_manager.py
```diff
@@ -8828,6 +8835,12 @@ def _patched_build_config(**kw):
                     # Claude strips CLAUDE_* from the PTY parent environment,
```
youchengsong, line 8934: **[MEDIUM] Resume-path cron suppression is untested**

Reviewer: `principal_engineer` · Category: `testing`

Evidence: The patch adds CLAUDE_DISABLE_CRON to the _patched_build_config resume/config overrides when task_id is present, but the added test only exercises im.launch() and inspects the direct subprocess environment. No test invokes or inspects _patched_build_config for a managed resumed session.

Impact: A future change could omit or misapply the override used when resuming an existing Claude session, allowing persisted provider-local cron schedules to revive despite the patch's stated lifecycle guarantee; the current regression suite would still pass.

Required fix: Add a focused test for the managed Claude _patched_build_config/resume path asserting CLAUDE_DISABLE_CRON=1 is present, and retain coverage that the override is absent for unmanaged sessions.

Verification: Exercise the patched config builder with provider='claude' and a non-null task_id, assert the generated overrides include CLAUDE_DISABLE_CRON='1', then repeat with task_id=None and assert it is not injected.

Finding fingerprint: `243d8302dcf25388128f78e171280ec6d68debb07b87247725548f0759255ea1`

youchengsong (commented):

The panel found 4 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The lifecycle changes are directionally coherent, but the patch introduces an uncontrolled external runtime dependency and does not prove that the safety-critical cron-disable switch is actually enforced by the resumed PTY implementation.

The in-repository lifecycle changes are coherent, but the runtime dependency is redirected from the project-controlled repository to a mutable branch in a contributor-owned fork, creating a blocking supply-chain and reproducibility risk.

The cron/runtime changes are covered by focused unit tests, but production now depends on a mutable personal fork branch for a runtime-critical package.

Reviewed commit: `d8db6e62064d399805ff65d1e2eadecb6d35e89c`.

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     # 分布式 Worker（EC2 provisioning）；WORKER_ENABLED=false 时不 import
     "boto3>=1.34.0",
     # PTY 模式（持久交互会话）；缺失时优雅降级为 `claude -p`
-    "claude-pty @ git+https://github.com/zjw49246/Claude-Code-PTY",
+    "claude-pty @ git+https://github.com/Citrus086/Claude-Code-PTY@fix/task-174-resume-merge",
```
youchengsong: **[HIGH] Runtime dependency points at a mutable personal fork branch**

Reviewer: `qa_engineer` · Category: `operations`

Evidence: The dependency was changed from the canonical `zjw49246/Claude-Code-PTY` repository to `Citrus086/Claude-Code-PTY@fix/task-174-resume-merge`; uv.lock likewise records the fork URL.

Impact: Fresh installs or lock regeneration depend on a contributor-owned branch that can be deleted, rewritten, or become unavailable, causing deployment failures or silently changing PTY behavior outside this repository's review and release controls.

Required fix: Merge the required PTY changes into the canonical repository (or an approved stable source) and pin an immutable commit there in both pyproject.toml and uv.lock.

Verification: Run a clean production-style `uv sync --locked`/build from the committed files and verify the package resolves from the approved canonical source and immutable revision.

Finding fingerprint: `cf59dd5c173005b601f7634f3af5c8644af5a89bcc740ed091910eab532c0b8c`

youchengsong: **[HIGH] Production dependency follows a mutable personal fork branch**

Reviewer: `principal_engineer` · Category: `operations`

Evidence: The dependency changes from the repository's upstream Claude-Code-PTY URL to `https://github.com/Citrus086/Claude-Code-PTY@fix/task-174-resume-merge`; only uv.lock records the current commit.

Impact: A lock refresh can silently select force-pushed or changed code, and deployments become dependent on a personal fork/branch remaining available. This is especially risky because the fork supplies the runtime behavior relied on for task isolation.

Required fix: Use a reviewed upstream release or an immutable commit in the dependency declaration itself (or move the required PTY changes into a controlled, versioned repository), then regenerate and verify the lockfile.

Verification: Run the dependency lock/update check from a clean checkout and assert that the resolved Claude-Code-PTY source and commit are immutable and match the declared requirement.

Finding fingerprint: `a589e70e2809e765810ec4f7edf4e731e23e0ab7466cf40458fe837491c67349`

backend/services/instance_manager.py
youchengsong: **[HIGH] Cron isolation relies on an unverified external environment contract**

Reviewer: `principal_engineer` · Category: `testing`

Evidence: The patch sets `CLAUDE_CODE_DISABLE_CRON=1` for managed Claude launches/resumed configuration, but the supplied tests only assert that the variable is present in the subprocess environment; no test demonstrates that the pinned PTY/Claude runtime consumes it or suppresses persisted schedules on resume.

Impact: If the external runtime ignores, renames, or regresses this switch, a resumed managed session can still revive a provider-local cron and generate orphan autonomous turns—the safety property this PR is intended to guarantee is then absent while tests remain green.

Required fix: Add a runtime-level contract/integration test covering a resumed managed PTY session with persisted cron state and asserting no cron turn is created, or replace the undocumented environment dependency with an explicit, verified PTY capability/API.

Verification: Construct a resumable managed session containing a native cron, launch it through the real PTY adapter, and verify the cron is disabled/removed and cannot emit a turn after the foreground task is terminalized.

Finding fingerprint: `ba03609206c6b42e137cea28eb5106374b94dfa74511237ab68df0c21a71034d`

pyproject.toml
youchengsong: **[MEDIUM] Runtime dependency trusts a mutable branch in a personal fork**

Reviewer: `senior_engineer` · Category: `security`

Evidence: The declared claude-pty dependency changes from github.com/zjw49246/Claude-Code-PTY to github.com/Citrus086/Claude-Code-PTY@fix/task-174-resume-merge. Although uv.lock currently records commit bd922ee85141b6c39cb78b5f9927cd008e1345f1, the authoritative pyproject declaration remains a movable branch, and installations or future lock regeneration can resolve different code controlled outside the project organization.

Impact: A force-push, account compromise, branch deletion, or ordinary branch update can silently alter or break a runtime dependency that handles PTY task execution. Environments resolving pyproject.toml without the existing lock are not reproducible from the reviewed head.

Required fix: Land the required claude-pty change in the project-controlled repository and reference an immutable reviewed revision in pyproject.toml, then regenerate uv.lock.

Verification: Perform a fresh locked dependency sync and lock consistency check, and run the resumed-turn regression against the immutable dependency revision.

Finding fingerprint: `e73961d71428212db6d12e7ac13d13baac3d3b901ddeeca5fbc929f5ec0f4382`

youchengsong (commented):

## Reviewer summaries

The runtime changes are cohesive, but the production dependency is redirected to a mutable personal fork reference without an immutable requirement in the project manifest.

The lifecycle changes are coherent, but the dependency switch introduces a mutable external code source that can change independently of this reviewed commit.

The in-process cleanup changes have regression coverage, but the core persisted-cron prevention relies on an external environment-variable contract that is not behaviorally verified.

Reviewed commit: `6c0257c81f7abf7a6008ecd3c536d1fc619bb8dc`.

backend/services/instance_manager.py
youchengsong: **[HIGH] Persisted-cron suppression is only tested as environment injection**

Evidence: The patch sets CLAUDE_CODE_DISABLE_CRON=1 for managed Claude launches and changes the PTY dependency to a fork, but the added test only inspects mock_exec.call_args.kwargs["env"]. No supplied test or in-repo implementation demonstrates that the selected Claude-Code-PTY/Claude runtime consumes this variable and suppresses schedules restored from an existing resumable session.

Impact: If the fork or provider ignores, renames, or changes the semantics of CLAUDE_CODE_DISABLE_CRON, a resumed managed task can still revive its persisted native cron and continue generating orphan autonomous turns—the primary safety behavior claimed by this PR.

Required fix: Pin/document a PTY/provider revision that implements this switch and add a behavioral integration test (using a fake or test PTY implementation) that starts a managed resumed session containing a persisted cron and verifies no cron/wakeup turn is created; cover both launch/resume paths that set the variable.

Verification: Run the new resumed-session test against the exact locked claude-pty revision and assert native CronCreate/CronList/ScheduleWakeup state is disabled, while CCM Monitor/Delivery scheduling remains functional.

Finding fingerprint: `b07f826de8d633caf97b3e8c9fb78eb49e5a8025489693bbdd0dea4bdbc89a2d`

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     # 分布式 Worker（EC2 provisioning）；WORKER_ENABLED=false 时不 import
     "boto3>=1.34.0",
     # PTY 模式（持久交互会话）；缺失时优雅降级为 `claude -p`
-    "claude-pty @ git+https://github.com/zjw49246/Claude-Code-PTY",
+    "claude-pty @ git+https://github.com/Citrus086/Claude-Code-PTY",
```
youchengsong: **[MEDIUM] Pin the PTY fork to an immutable revision**

Reviewer: `senior_engineer` · Category: `operations`

Evidence: The dependency is changed to `claude-pty @ git+https://github.com/Citrus086/Claude-Code-PTY` with no tag or commit in pyproject.toml. Although uv.lock currently records commit bd922ee..., installs that do not use the lock (or a lock refresh) will resolve the mutable fork default branch, whose contents are outside this reviewed head.

Impact: A future fork update can silently alter PTY resume and cron-disable behavior, defeating reproducibility and potentially reintroducing the orphan-turn lifecycle bug or other security/regression issues without a corresponding CCM review.

Required fix: Reference the exact reviewed PTY commit (or an immutable release/tag) in pyproject.toml and keep uv.lock consistent; enforce frozen-lockfile installation in deployment/CI.

Verification: From a clean environment, run the project's frozen dependency installation and verify the resolved claude-pty source is exactly the pinned commit; a lock-refresh check should fail if pyproject.toml and uv.lock diverge.

Finding fingerprint: `18c32f225bd82b7d4d40c6f05f14723d262b319c2ef377531a54a8d4f8972d45`

youchengsong: **[MEDIUM] Pin the external Claude PTY fork immutably**

Reviewer: `principal_engineer` · Category: `operations`

Evidence: The dependency changes from the repository-owned URL to `claude-pty @ git+https://github.com/Citrus086/Claude-Code-PTY` with no commit or tag selector. Although `uv.lock` currently records commit `bd922ee85141b6c39cb78b5f9927cd008e1345f1`, regenerating the lock or installing through tooling that resolves the manifest can silently select whatever the fork's default branch points to.

Impact: Production environments can receive unreviewed or incompatible PTY code, causing behavior drift from the captured and tested revision and expanding the supply-chain trust boundary to a mutable personal branch.

Required fix: Reference a reviewed immutable commit (or signed release/tag) directly in `pyproject.toml`, and keep `uv.lock` synchronized to that exact revision.

Verification: Run `uv lock --check` and perform a clean dependency installation in CI, asserting that the resolved `claude-pty` source URL and commit equal the reviewed immutable revision.

Finding fingerprint: `937492a44a1df523fb53bb3db519cf067d8b657d78c3a5942efa48eef5926d18`

youchengsong (commented):

The panel found 4 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The lifecycle changes are directionally coherent, but the patch moves a production dependency to an independently controlled mutable fork without an immutable requirement in project metadata.

核心隔离与状态清理逻辑合理，但新的 PTY 依赖在清单中跟随个人 fork 的可变默认分支，形成阻塞性的供应链风险。

The configuration paths are unit-tested, but the two provider-boundary behaviors on which the fix depends lack behavioral regression coverage.

Reviewed commit: `1d582c901749ef28b177fed9101637df9f3d1b0c`.

pyproject.toml
youchengsong: **[HIGH] Production dependency now trusts a mutable personal fork**

Reviewer: `principal_engineer` · Category: `security`

Evidence: The dependency was changed from `https://github.com/zjw49246/Claude-Code-PTY` to `https://github.com/Citrus086/Claude-Code-PTY` with no commit or tag pin. Although uv.lock records a commit, the project requirement itself resolves the fork's default branch whenever the lock is regenerated or bypassed.

Impact: A repository outside the project's normal owner/review boundary can change PTY runtime code and have those changes enter deployments through dependency resolution, undermining reproducibility and creating a supply-chain risk in the process responsible for managed task isolation.

Required fix: Use an immutable, reviewed commit or release tag in `pyproject.toml` (preferably after merging the required PTY changes into the canonical dependency repository), and regenerate/validate `uv.lock` from that exact reference.

Verification: Run dependency resolution in a clean environment with lock enforcement and without lock enforcement; verify both select the same reviewed PTY commit and fail if the reference changes.

Finding fingerprint: `c0b154f05bd130426d0cf9c07956dfaff015631871b6c0591f23853d3b2cae76`

backend/services/instance_manager.py
youchengsong: **[MEDIUM] Persisted-cron prevention is not behaviorally verified**

Evidence: The new test replaces create_subprocess_exec with a mock and only asserts that CLAUDE_CODE_DISABLE_CRON appears in the constructed environment. It never resumes a session containing an existing native cron schedule or verifies that no autonomous turn is produced. The disallowed-tools test likewise covers only command construction and cannot prove the legacy-session path claimed by the PR.

Impact: If the runtime does not honor this variable, changes its contract, or loses it before session restoration, managed tasks created before this restriction can still revive orphan cron turns while all added tests pass.

Required fix: Add a provider-boundary regression test, using the supported Claude test harness or a pinned runtime fixture, that resumes a session with persisted cron state under the generated managed-task configuration and proves the schedule is disabled and produces no turn.

Verification: Create persisted native cron state, resume it through the managed-task launch path, advance or wait through its due time, and assert that no autonomous message/turn is emitted; include a control showing the fixture emits one without the disable setting.

Finding fingerprint: `1f365a492bc2e4124491b2ebcc46bfc54c983a2bb6505e63eb2e673248ea81d7`

uv.lock
```diff
@@ -519,7 +519,7 @@ dev = [
 [[package]]
 name = "claude-pty"
 version = "0.1.0"
-source = { git = "https://github.com/zjw49246/Claude-Code-PTY#f643019787d1c8f8d0454589323c82bfa8905426" }
+source = { git = "https://github.com/Citrus086/Claude-Code-PTY#32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f" }
```
youchengsong: **[MEDIUM] Queued-resume runtime upgrade has no regression test**

Evidence: The final dependency change moves claude-pty from commit bd922ee85141b6c39cb78b5f9927cd008e1345f1 to 32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f specifically for queued resume alignment, but none of the supplied test changes exercises a queued message during PTY session resume; the added tests cover command environment construction and stale-state reconciliation instead.

Impact: The opaque dependency update can still drop, duplicate, or misorder a queued user turn during resume, and the repository test suite shown here would not detect the regression.

Required fix: Add an integration regression around the installed claude-pty interface that queues a user turn while a prior session is being resumed and verifies exact-once delivery in the correct resumed session.

Verification: Start a resumable PTY session, initiate resume, enqueue a message during the resume window, and assert the resumed transcript contains that message exactly once and after the restored history.

Finding fingerprint: `161c683167a4bcc4034071ad2a1bfda10863d31ff43368d9e882f08b1f450c40`

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
...
     "pyjwt>=2.13.0",
```
youchengsong, line 25: **[MEDIUM] Executable dependency follows a contributor fork's mutable default branch**

Reviewer: `senior_engineer` · Category: `security`

Evidence: The manifest changes `claude-pty` from the repository-owner namespace to `git+https://github.com/Citrus086/Claude-Code-PTY` without a revision, while `uv.lock` happens to resolve commit `32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f`. Thus the manifest delegates future dependency resolution to the current HEAD of an externally controlled default branch rather than identifying the reviewed code.

Impact: A routine lock regeneration or dependency upgrade can silently select new executable backend code from that fork. Compromise, force-push, or unrelated changes in the fork can therefore enter the application without an explicit dependency revision change in the manifest; the lock only protects installations using the current frozen lock.

Required fix: Pin the direct requirement to the reviewed immutable commit (for example, append `@32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f`) or land that commit in an organization-controlled upstream and reference it there, then regenerate the lock.

Verification: Run the repository's frozen dependency validation and assert that `pyproject.toml` and `uv.lock` both resolve `claude-pty` to the same immutable commit; regenerating the lock must leave that commit unchanged.

Finding fingerprint: `a093df4e495af5342c3a6757fd131103931f3a28c142872bfb91b7e31ad9ebd0`

youchengsong (commented):

The panel found 5 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The runtime-guard recovery has an incomplete cleanup path, and the PTY backend is moved from the project-owned repository to a contributor-owned moving branch without reviewable proof or durable ownership.

The runtime-guard reconciliation can admit a new turn while a live PTY runtime remains, and the dependency manifest introduces an unpinned executable dependency from a contributor-owned fork.

Ownerless retryable Tasks can retain dead PTY guards and remain permanently blocked from redispatch.

Reviewed commit: `9d1c626a60eb4db75cf345933f28b13234ec36ae`.

backend/services/dispatcher.py
```diff
@@ -22918,6 +22942,16 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
             return True
         session_id = task.session_id
         if session_id:
+            reconcile_runtime_guard = getattr(
+                self.instance_manager,
+                "reconcile_orphaned_pty_runtime_guard",
+                None,
+            )
+            terminal_ownerless = bool(
```
youchengsong, line 22966: **[MEDIUM] Pending Tasks cannot reconcile orphaned PTY runtime guards**

Evidence: The stale-state cleanup added in this PR can release a Task to either `pending` or `failed` and clears `pty_background_generation` for both states. However, `_queued_task_has_live_generation` only calls `reconcile_orphaned_pty_runtime_guard` when an ownerless Task has a terminal status (`completed`, `failed`, `cancelled`, or `conflict`). An ownerless Task reset to `pending` with a dead in-memory background generation or handoff therefore returns `True` without attempting reconciliation.

Impact: A retryable Task recovered to `pending` can remain indefinitely classified as having a live generation, preventing redispatch and reproducing the stranded-task behavior this PR is intended to fix.

Required fix: Permit orphan-guard reconciliation for ownerless `pending` Tasks as well, relying on `reconcile_orphaned_pty_runtime_guard`'s live-session and callback-owner checks to retain genuinely active guards.

Verification: Create an ownerless `pending` Task with a real dead `_PtyBackgroundState` or autonomous handoff in `InstanceManager`, call `_queued_task_has_live_generation`, and assert that the production reconciler removes the guard and the method returns `False`; also verify that a live session still returns `True`.

Finding fingerprint: `9f38a3c5f8b11a57b1af6f9ed23d658486cfb9e3d5e1421188b581e1af9cb0bf`

backend/services/instance_manager.py
```diff
@@ -9966,6 +9979,69 @@ def reset_pty_autonomous_activity_handoff(
...
+            ):
+                return False
+
+            proof = self._pty_post_exit_generations.get(key)
+            if (
+                proof is not None
+                and getattr(proof.session, "is_alive", True) is not False
+            ):
+                return False
+            runtime_sessions = self._task_pty_runtime_session_candidates(
```
youchengsong, line 10122: **[MEDIUM] Reconciliation can return success while retaining a post-exit guard**

Evidence: The method reads `_pty_post_exit_generations[key]` and permits cleanup when its session is dead, but only discards the background state and calls `reset_pty_autonomous_activity_handoff()`. The shown reset implementation discards a post-exit generation only when a handoff token was present. Therefore a dead proof with no handoff remains stored even though the method returns `True`, contradicting its contract that no runtime guard remains. The added tests cover a handoff-only case and a background-state-plus-handoff case, not this path.

Impact: The dispatcher may admit a new queued turn after being told reconciliation completed while stale per-task/session runtime state remains registered, undermining the transition-lock ownership invariant and allowing later callback or cleanup logic to observe an obsolete generation.

Required fix: While holding `pty_background_transition`, explicitly discard the exact dead post-exit proof when it exists, independently of whether a handoff token exists; alternatively return `False` until that proof is safely retired.

Verification: Populate `_pty_post_exit_generations` with a dead-session proof and no autonomous handoff, call `reconcile_orphaned_pty_runtime_guard`, and assert that success implies the proof, background generation, and handoff are all absent. Also cover the same proof alongside a dead background state.

Finding fingerprint: `64970f2b9e6a4d409375522469c45f739ee16bd6567454cc57415a5d634d15cd`

pyproject.toml
youchengsong: **[MEDIUM] Core PTY backend is redirected to a contributor-owned floating repository**

Reviewer: `principal_engineer` · Category: `operations`

Evidence: The dependency source changes from `github.com/zjw49246/Claude-Code-PTY` to `github.com/Citrus086/Claude-Code-PTY` without a revision in `pyproject.toml`. Although this lockfile snapshot records commit `32f1e5ac...`, future lock refreshes resolve the contributor fork's moving default branch. The external implementation change described as resumed-turn alignment is not present in the supplied review material.

Impact: A correctness-critical runtime backend and future executable dependency updates are now controlled outside the project's repository ownership, making reproducibility after lock refresh, incident rollback, and maintenance contingent on a personal fork. The panel also cannot verify the backend behavior on which this PR relies.

Required fix: Land the required PTY commits in a project-controlled repository and reference an immutable reviewed revision or release from that repository, then regenerate the lockfile.

Verification: Resolve/install from a clean environment using the committed lockfile and add an integration regression proving queued/resumed PTY turns remain aligned across disconnect and resume with the pinned backend revision.

Finding fingerprint: `a9fe122f555cee33da48ed42ee2e7dd2439426475aeed0067899c97f471523d7`

backend/services/instance_manager.py
```diff
@@ -9966,6 +9979,69 @@ def reset_pty_autonomous_activity_handoff(
...
+
+        Return ``True`` when no runtime guard remains for the Task/session.
+        """
+
+        key = (task_id, session_id)
+        async with self.pty_background_transition(task_id, session_id):
+            state = self._pty_background_states.get(key)
+            has_handoff = key in self._pty_autonomous_activity_handoffs
+            if state is None and not has_handoff:
+                return True
```
youchengsong: **[MEDIUM] Guard reconciliation skips live-runtime checks after state removal**

Reviewer: `senior_engineer` · Category: `concurrency`

Evidence: Inside `pty_background_transition`, `reconcile_orphaned_pty_runtime_guard` returns `True` immediately when `_pty_background_states` and `_pty_autonomous_activity_handoffs` contain no entry. This return occurs before checking active handoff owners, `_pty_post_exit_generations`, or `_task_pty_runtime_session_candidates`, despite the method contract saying live sessions and post-exit proofs must fail closed. The dispatcher interprets `True` as permission to continue and, for an ownerless terminal task, can return `False` from `_queued_task_has_live_generation`.

Impact: If the initially observed background state or handoff is removed while reconciliation waits for the transition lock, but its session, callback owner, or post-exit proof remains live, a queued turn can launch concurrently with that PTY runtime.

Required fix: Perform all live owner, post-exit proof, and runtime-session checks before returning success for the no-state/no-handoff case. Return `True` only after confirming that none of those runtime guards is live.

Verification: Arrange for the initial generation or handoff lookup to trigger reconciliation, then have the transition-lock-protected state disappear while a live post-exit proof or runtime-session candidate remains. Assert reconciliation returns `False` and `_queued_task_has_live_generation` remains `True`.

Finding fingerprint: `d11caa935546dc83088c84092c22be438594b56af46ddddb4fd23b9a29377a93`

pyproject.toml
youchengsong: **[MEDIUM] Executable dependency is redirected to an unpinned contributor fork**

Reviewer: `senior_engineer` · Category: `security`

Evidence: The manifest changes `claude-pty` from the project-owned repository to `git+https://github.com/Citrus086/Claude-Code-PTY` without a revision. Although the current lock records commit `32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f`, installations or lock regeneration from `pyproject.toml` resolve the mutable default branch under a different trust owner.

Impact: A later branch update can execute dependency code that was not represented by this reviewed head, making builds non-reproducible and expanding the supply-chain trust boundary.

Required fix: Pin the direct dependency URL in `pyproject.toml` to the reviewed commit, or move the required changes to a repository and immutable release controlled by the project; regenerate the lock accordingly.

Verification: Resolve dependencies from the manifest without the existing lock and assert the resulting `claude-pty` source revision is the explicitly approved immutable commit.

Finding fingerprint: `e9a527e0e9955e2830628597a389ae38176b629664c93c33757b492d245c8ef8`

youchengsong (commented):

The panel found 4 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The runtime changes are structurally plausible, but the PR moves an executable dependency to an unpinned contributor-owned branch, creating a blocking supply-chain and reproducibility regression.

The runtime-guard reconciliation has a race that can admit a new turn while an old PTY runtime remains live.

The tool deny-list is covered, but the runtime cron safeguard has an untested critical launch path, and the PTY dependency now trusts a mutable personal-fork branch.

Reviewed commit: `e0ce251da1a3c87802ea768ddaf100e466123fdf`.

backend/services/instance_manager.py
youchengsong, line 8934: **[MEDIUM] Cron-disable override is untested on the path that requires it**

Evidence: The patch adds CLAUDE_CODE_DISABLE_CRON to `_patched_build_config` overrides because, according to the adjacent comment, Claude strips CLAUDE_* variables from the PTY parent environment. The added `test_managed_claude_launch_disables_native_cron` only intercepts `asyncio.create_subprocess_exec` and asserts its parent `env`; no supplied test exercises the patched PTY config builder or proves the explicit override reaches the Claude process.

Impact: The subprocess-env test can pass while managed PTY/resume sessions still omit the switch and revive persisted native cron schedules, defeating the PR's primary lifecycle guarantee in the exact path where normal environment propagation is known to be insufficient.

Required fix: Add regression coverage for the PTY wrapper/config construction path that asserts `CLAUDE_CODE_DISABLE_CRON=1` is present in the final explicit overrides for managed Claude tasks and absent where appropriate for unmanaged sessions.

Verification: Build or launch a managed Claude PTY session through `_patched_build_config` using a fake backend/wrapper, capture the final config passed to the provider process, and assert the cron-disable override is present; include a resumed-session case because persisted schedules are the stated risk.

Finding fingerprint: `78d1b3bf4e629ccc4090bb93f3ce91ddaf576ec1c442850791b042bb0de5c5be`

pyproject.toml
youchengsong: **[MEDIUM] Production dependency follows a mutable personal-fork branch**

Reviewer: `qa_engineer` · Category: `security`

Evidence: The direct dependency changes from the project repository to `git+https://github.com/Citrus086/Claude-Code-PTY` without an immutable revision. Although `uv.lock` currently records commit `32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f`, installers resolving `pyproject.toml` directly can fetch whatever the fork's default branch contains later.

Impact: Equivalent builds can install different or unreviewed PTY runtime code, and compromise or force-push of the contributor-owned fork can affect production installations. This dependency controls session/resume behavior central to the fix.

Required fix: Reference an immutable reviewed commit in `pyproject.toml`, preferably from a repository under the project's control, and regenerate the lock so both declarations identify the same commit.

Verification: In a clean environment, resolve/install from `pyproject.toml` without the existing lock and assert the resolved `claude-pty` source commit equals the reviewed pinned SHA.

Finding fingerprint: `f3d15a959adb2febf5698c42b15ef4fc476af091aa73dfd2c43b21ac2a6f5ac7`

backend/services/instance_manager.py
```diff
@@ -9966,6 +9979,69 @@ def reset_pty_autonomous_activity_handoff(
...
+        A forced PTY timeout can stop the native Session and fail-close the
+        durable Task while a background state and/or autonomous handoff token
+        remains indexed.  Those guards no longer have a durable owner, so a
+        later chat turn would otherwise wait forever.  Serialize with the
+        callback admission lock and keep any live Session, callback owner, or
+        post-exit proof fail-closed.
+
+        Return ``True`` when no runtime guard remains for the Task/session.
+        """
+
```
youchengsong, line 10098: **[MEDIUM] Early return bypasses live runtime-owner checks**

Reviewer: `senior_engineer` · Category: `concurrency`

Evidence: Inside reconcile_orphaned_pty_runtime_guard(), `if state is None and not has_handoff: return True` runs before checking unfinished handoff owners, `_pty_post_exit_generations`, and `_task_pty_runtime_session_candidates()`. The dispatcher invokes this after observing a guard but before acquiring the transition lock; while it waits, another transition can remove that observed state or handoff. The method can then return True solely because those two indexes are empty, even though a callback owner, post-exit proof, or backend/pool session is still alive. The added tests cover a retained handoff owner and a dead background state, but not this disappearing-index/live-runtime case.

Impact: `_queued_task_has_live_generation()` can treat a terminal ownerless task as idle and launch its queued message while the previous PTY session or callback still owns the task, recreating competing turns and violating the runtime-ownership invariant this patch is intended to restore.

Required fix: Do not return successfully based only on absence of background state and handoff. Perform the owner, post-exit proof, and runtime-session liveness checks first; return True only after all exact runtime-owner sources are absent or dead.

Verification: Arrange for the initial dispatcher lookup to observe a guard, then have that guard disappear before reconciliation acquires the transition lock while `_task_pty_runtime_session_candidates()` still contains an `is_alive=True` session (and separately while an unfinished handoff owner remains). Assert reconciliation returns False and the dispatcher continues to report a live generation.

Finding fingerprint: `c267ca63a259d367d90e5e286cd81d71b118a55d549d4dfd217cabb66e64e490`

pyproject.toml
youchengsong: **[HIGH] Executable PTY dependency now tracks a mutable contributor fork**

Reviewer: `principal_engineer` · Category: `security`

Evidence: The dependency changes from the repository-owner namespace `github.com/zjw49246/Claude-Code-PTY` to `github.com/Citrus086/Claude-Code-PTY` without an immutable revision in `pyproject.toml`. Although this captured `uv.lock` resolves commit `32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f`, installers resolving directly from project metadata and future lock regeneration will fetch whatever the fork's default branch contains.

Impact: Dependency installation can silently execute code different from the reviewed head, and control of a contributor account or mutable branch becomes control of backend runtime code. Builds that do not enforce the captured lock are not reproducible, and rollback cannot reliably restore the reviewed dependency behavior.

Required fix: Reference an approved repository at an immutable audited commit in `pyproject.toml` and regenerate `uv.lock` so both files identify the same source and commit. If the fork must remain, pin `32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f` explicitly rather than its default branch.

Verification: In a clean environment, resolve dependencies from `pyproject.toml`, verify the resulting `claude-pty` source commit is exactly the approved SHA, and run the PTY resume and queued-resume regression tests against that resolved artifact.

Finding fingerprint: `f6c6aa1f6a2fc13a58d8e44a3500b7a96b911d1a8add2a0a54fd5d29f0673273`

youchengsong (commented):

The panel found 4 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The lifecycle changes are broadly cohesive, but the runtime dependency is redirected to a mutable personal fork, creating an avoidable reproducibility and supply-chain boundary.

The stale-runtime recovery has two concurrency gaps: superseded tasks are not reconciled, and the lifecycle heuristic can cancel a legitimate launch before its durable state is recorded.

The cron restrictions are covered at command/environment boundaries, but stale lifecycle reconciliation can admit a replacement before the cancelled lifecycle has actually stopped.

Reviewed commit: `e9e5e95bc33e3057b56bb4d049104016693c5f2e`.

pyproject.toml
youchengsong: **[MEDIUM] Do not depend on an unpinned personal fork branch**

Reviewer: `principal_engineer` · Category: `operations`

Evidence: The patch changes the claude-pty requirement from the repository-owned URL to `git+https://github.com/Citrus086/Claude-Code-PTY` without a tag or commit pin; only the separately generated uv.lock records a commit.

Impact: Any lock regeneration or installation path that does not strictly honor the existing lock can silently consume future changes from a third-party `main` branch, making deployments non-reproducible and expanding the trusted code boundary for PTY session handling.

Required fix: Use an immutable commit (or an owned, reviewed upstream release) in pyproject.toml and keep uv.lock aligned; document the fork/provenance if the fork is unavoidable.

Verification: Run dependency resolution from a clean checkout and verify the resolved claude-pty source is the intended immutable commit; modify the remote branch and confirm resolution cannot change without an explicit dependency update.

Finding fingerprint: `f0cb9989b6c7ffb13bed3d858eb2a89a1e9c687095ef28b37af075ea143e477f`

backend/services/dispatcher.py
```diff
@@ -22957,9 +23009,43 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
             return False
 
         lifecycle = self._running_tasks.get(instance_id)
+        manager_running = bool(self.instance_manager.is_running(instance_id))
         if lifecycle is not None and not lifecycle.done():
             lifecycle_task_id = getattr(lifecycle, "_ccm_task_id", None)
```
youchengsong, line 23042: **[HIGH] Stale-lifecycle heuristic can cancel a fresh launch**

Reviewer: `senior_engineer` · Category: `concurrency`

Evidence: The new branch cancels any lifecycle bound to the task when the durable task is terminal and the instance still appears idle/error and ownerless. The immediately following existing comment states that fresh lifecycle preparation precedes persistence of Instance.current_task_id/PID and that the in-memory binding is authoritative during that launch window. A newly registered lifecycle for a retry can therefore satisfy the new stale conditions before its durable launch fields are written and be cancelled as stale.

Impact: A concurrent queued-message check can cancel a valid replacement turn during launch admission, dropping or indefinitely delaying the user's message. Detaching the lifecycle before cancellation also allows another launch to be admitted while cancellation cleanup is still running.

Required fix: Give lifecycle registrations an exact generation or explicit launch/cleanup phase and only retire a lifecycle proven to belong to an already-released generation. Do not infer staleness solely from the temporarily unpersisted instance snapshot; complete or await cancellation under the relevant admission barrier before reporting the task as free.

Verification: Pause a real/fake launch after registering `_running_tasks` but before persisting instance ownership/PID, invoke `_queued_task_has_live_generation`, and assert the fresh lifecycle remains registered and uncancelled. Separately prove that a lifecycle from an older released generation is cancelled and fully quiesced before a replacement is admitted.

Finding fingerprint: `366f74a6ff9aba1a22e0904595698a2b2c1e758c24ceb47e6a9c6317c85ab6e8`

backend/services/dispatcher.py
```diff
@@ -22918,6 +22942,16 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
             return True
         session_id = task.session_id
         if session_id:
+            reconcile_runtime_guard = getattr(
+                self.instance_manager,
+                "reconcile_orphaned_pty_runtime_guard",
+                None,
```
youchengsong, line 22964: **[MEDIUM] Superseded tasks cannot retire orphaned PTY guards**

Evidence: `terminal_ownerless` permits reconciliation only for completed, failed, cancelled, and conflict states. The later stale-lifecycle branch explicitly treats `superseded` as terminal, but it is omitted here. Consequently a superseded ownerless task with a dead background generation or handoff returns `True` before reaching the ownerless fallback and never calls `reconcile_orphaned_pty_runtime_guard`.

Impact: A dead PTY guard attached to a superseded task can continue making `_queued_task_has_live_generation` report busy, permanently blocking a later queued chat turn.

Required fix: Include `superseded` in the terminal states eligible for orphaned PTY guard reconciliation, preferably by using one authoritative terminal-status set for both branches.

Verification: Repeat both orphaned-handoff and dead-background-state reconciliation tests with an ownerless task whose status is `superseded`, and assert the guard is removed and the method returns false.

Finding fingerprint: `d7497a75896ac3b10fd2b49395c65225e64cadcf6cef9bedf8b288252c5516b9`

backend/services/dispatcher.py
```diff
@@ -22957,9 +23009,43 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
...
+                    and instance.current_task_id is None
+                    and instance.current_plan_run_id is None
+                    and instance.pid is None
+                    and not manager_running
+                ):
+                    # A lifecycle can remain suspended in terminal cleanup
+                    # after its exact process/consumer and reverse owner have
+                    # already been released.  It has no remaining execution
+                    # authority, so retaining it here would block every later
+                    # chat turn forever.  Detach and cancel only from this
```
youchengsong: **[MEDIUM] Cancelled lifecycle is treated as stopped before cancellation completes**

Reviewer: `qa_engineer` · Category: `concurrency`

Evidence: The new terminal-ownerless branch removes the lifecycle from `_running_tasks`, calls `lifecycle.cancel()`, and immediately returns `False`. In asyncio, `cancel()` only requests cancellation; the coroutine and its `finally` cleanup can continue running until it reaches cancellation and completes. The added test only checks `lifecycle.cancelling()` and awaits it after `_queued_task_has_live_generation` has already returned, so it does not prove replacement admission waits for termination.

Impact: The caller can launch a later chat turn while the old lifecycle is still executing cleanup against the same task or instance. This can race session teardown, state persistence, or other cleanup with the replacement generation despite the method reporting that no live generation exists.

Required fix: Do not return `False` until the cancelled lifecycle is done. Either await its completion with the existing bounded lifecycle-cancellation mechanism and fail closed on timeout, or keep reporting the task busy until a completion callback removes the lifecycle and wakes dispatch.

Verification: Use a lifecycle coroutine whose cancellation handler blocks on an event. Verify `_queued_task_has_live_generation` cannot report `False` while that handler is blocked, then release it and verify reconciliation permits the queued turn; also cover the timeout path remaining fail-closed.

Finding fingerprint: `a782281a4a2a6a1f2bb9babe6e232353cc55fe8222b86846c8caff2ccf7ec42c`

youchengsong (commented):

The panel found 4 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The cron isolation is directionally sound, but the patch creates a dependency trust-boundary regression and leaves one recognized terminal state outside PTY guard recovery.

The orphaned PTY guard recovery omits one terminal task state, allowing a dead guard to keep later messages queued indefinitely.

The cron restrictions are covered, but PTY guard reconciliation can incorrectly admit a new turn while a live runtime proof remains.

Reviewed commit: `8c369010db077c6e506ee0b2879d4949008c1037`.

backend/services/dispatcher.py
```diff
@@ -22916,8 +22940,38 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
...
+        )
         session_id = task.session_id
         if session_id:
+            reconcile_runtime_guard = getattr(
+                self.instance_manager,
+                "reconcile_orphaned_pty_runtime_guard",
+                None,
+            )
+            terminal_ownerless = bool(
+                task.status
```
youchengsong, line 22967: **[MEDIUM] Superseded tasks cannot retire orphaned PTY runtime guards**

Reviewer: `senior_engineer` · Category: `concurrency`

Evidence: The new terminal_ownerless predicate permits reconciliation only for completed, failed, cancelled, and conflict tasks. The same method's stale-lifecycle branch explicitly treats superseded as terminal, confirming it is a valid terminal state. With status superseded, a retained dead background generation or handoff therefore bypasses reconcile_orphaned_pty_runtime_guard and is returned as live. The added guard tests cover failed tasks only.

Impact: A superseded task carrying a stale PTY background state or autonomous handoff can remain permanently busy, so a later queued chat turn never launches even though its process, instance ownership, and manager runtime are gone.

Required fix: Include superseded in the terminal_ownerless status set used for PTY runtime-guard reconciliation.

Verification: Add a _queued_task_has_live_generation regression test using a superseded task, an ownerless idle/error instance, and a stale handoff or background generation; assert reconciliation is awaited and the method returns false after the guard is removed.

Finding fingerprint: `fb8b1aadcfbecd025d5bb454d7b501324ce9e7c6a8d5b4b26bce6cd037fe8370`

backend/services/instance_manager.py
```diff
@@ -9966,6 +9979,69 @@ def reset_pty_autonomous_activity_handoff(
...
+        key = (task_id, session_id)
+        async with self.pty_background_transition(task_id, session_id):
+            state = self._pty_background_states.get(key)
+            has_handoff = key in self._pty_autonomous_activity_handoffs
```
youchengsong, line 10102: **[HIGH] Early return bypasses live PTY runtime and post-exit checks**

Reviewer: `qa_engineer` · Category: `concurrency`

Evidence: reconcile_orphaned_pty_runtime_guard returns True immediately when both the background state and handoff are absent. This occurs before checking _pty_post_exit_generations and _task_pty_runtime_session_candidates, despite the method contract stating that live post-exit proofs and Sessions must remain fail-closed. The caller observes a guard before awaiting this method, so another transition can remove that guard while it waits for pty_background_transition yet leave a live post-exit proof or runtime Session; the early return then reports reconciliation success. The added tests only cover a live handoff owner and a dead background state, not this state transition.

Impact: A queued message can launch against a resumable Claude session that still has a live PTY runtime or callback proof, creating concurrent writers and recreating the autonomous-turn/user-message competition this PR is intended to prevent.

Required fix: Do not return solely because state and handoff are absent. Check handoff owners, post-exit proof, and runtime-session candidates first, and return True only after all runtime evidence is absent or dead.

Verification: Add a unit test with no background state or handoff but a live _pty_post_exit_generations entry, and another with a live runtime-session candidate; each must assert reconciliation returns False and preserves the proof/session. Also cover the race-shaped case where the initially observed guard disappears before the transition lock is acquired.

Finding fingerprint: `ba47ba34772a84df1d57aa8d1b9f794b0ff078d1dd304bdd1b780eb675dd3e28`

pyproject.toml
youchengsong: **[HIGH] Runtime dependency is redirected to a mutable contributor-owned fork**

Reviewer: `principal_engineer` · Category: `security`

Evidence: The dependency source changes from `github.com/zjw49246/Claude-Code-PTY` to `github.com/Citrus086/Claude-Code-PTY` without an immutable revision in `pyproject.toml`. Although `uv.lock` records commit `32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f`, installers or lock regeneration that resolve directly from project metadata execute whatever the fork's default branch contains at that time.

Impact: A contributor-controlled repository becomes an executable production dependency and can change independently of this repository's review and release process, compromising reproducibility and expanding the software supply-chain trust boundary.

Required fix: Use the organization-maintained dependency after upstreaming the required commits, or pin an explicitly reviewed immutable commit under an organization-controlled repository in both `pyproject.toml` and `uv.lock`.

Verification: In a clean environment, resolve/install from `pyproject.toml` without reusing the existing lock and assert that the resolved `claude-pty` source is the approved repository and exact reviewed commit.

Finding fingerprint: `8a997481e00455fcf4080c4cdc6b55c7bfa5b3b1972f95f7517ab2d85d7e24d9`

backend/services/dispatcher.py
```diff
@@ -22916,8 +22940,38 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
...
+            and self.instance_manager.is_running(instance_id)
+        )
         session_id = task.session_id
         if session_id:
+            reconcile_runtime_guard = getattr(
+                self.instance_manager,
+                "reconcile_orphaned_pty_runtime_guard",
+                None,
+            )
+            terminal_ownerless = bool(
```
youchengsong, line 22966: **[MEDIUM] Superseded tasks cannot retire orphaned PTY runtime guards**

Reviewer: `principal_engineer` · Category: `correctness`

Evidence: `terminal_ownerless` permits reconciliation only for `completed`, `failed`, `cancelled`, and `conflict`. The lifecycle reconciliation immediately below treats `superseded` as a terminal status as well, establishing that it is part of the same terminal-state model. A superseded task with a stale background generation or handoff therefore returns `True` before reaching lifecycle cleanup.

Impact: A stale PTY guard associated with a superseded task can keep later queued chat pending indefinitely even when its instance is idle/error, pid-less, ownerless, and no manager runtime remains.

Required fix: Include `superseded` in the terminal statuses eligible for orphaned PTY guard reconciliation, preferably by reusing one authoritative terminal-status set for both guard and lifecycle recovery.

Verification: Add the retained-dead-instance guard reconciliation test with task status `superseded`; verify `_queued_task_has_live_generation` returns false and `reconcile_orphaned_pty_runtime_guard` is awaited.

Finding fingerprint: `d1d3ed118b0966e41ee289f46670d6935f159a4edc308408c2bf7539570b82b0`

youchengsong (commented):

The panel found 4 open blocking findings on this reviewed PR head. Address the inline finding threads before merging.

## Reviewer summaries

The lifecycle changes are broadly coherent, but stale PTY routing markers and the unreviewed dependency-origin change leave concrete operational risks.

Stale lifecycle reconciliation can never release real terminal lifecycles because it compares mutable completion timestamps against the launch snapshot.

The stale-lifecycle recovery compares a launch-time snapshot against a terminal timestamp, so the production transition it is meant to recover can remain blocked indefinitely; the regression test does not reproduce that transition.

Reviewed commit: `1c260a77f120a2b1d2c79b5893bb7006457e5333`.

backend/services/dispatcher.py
```diff
@@ -22961,6 +23041,89 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
...
+                    and instance.status in {"idle", "error"}
+                    and instance.current_task_id is None
+                    and instance.current_plan_run_id is None
+                    and instance.pid is None
+                    and not manager_running
+                ):
+                    lifecycle_generation = getattr(
+                        lifecycle,
+                        "_ccm_task_generation",
+                        None,
```
youchengsong, line 23063: **[HIGH] Mutable terminal timestamps prevent stale lifecycle cleanup**

Reviewer: `senior_engineer` · Category: `concurrency`

Evidence: The new reconciliation requires the stored `_ccm_task_generation` to match the current Task across `started_at` and `completed_at`. The lifecycle generation is captured when the task is launched, while terminal cleanup subsequently sets `completed_at` (and may update other lifecycle timestamps), so a genuinely ownerless terminal lifecycle will fail this comparison and return `True` instead of being cancelled and removed. The added test constructs the lifecycle generation only after creating an already-failed task, so it does not exercise this production transition.

Impact: After a PTY/process failure, the stale lifecycle remains in `_running_tasks` and continues to make later queued turns appear busy indefinitely, defeating the fix for the orphaned lifecycle race.

Required fix: Use only immutable task-generation identity fields for this comparison, or persist/capture a terminal generation snapshot that is intentionally unchanged by completion bookkeeping; do not compare mutable `started_at`/`completed_at` values.

Verification: Register a lifecycle while the task is executing, then transition the same task to a terminal status and set `completed_at`; assert `_queued_task_has_live_generation` cancels/removes the exact lifecycle and returns `False`, while a lifecycle for a different generation remains busy.

Finding fingerprint: `6290c9e63f2e9338d4de193193a121b7edcae4af35f3c4e081cad1b28444c5c0`

backend/services/dispatcher.py
```diff
@@ -4128,6 +4130,13 @@ async def _cleanup_stale_state(
                         t.id,
                         t.status,
                     )
+                if new_status in {"pending", "failed"}:
+                    # Once the exact runtime has been released or failed
+                    # closed, no retained PTY tail can still accept input.
```
youchengsong, line 4135: **[MEDIUM] Terminal PTY routing markers are not cleared for every terminal state**

Reviewer: `principal_engineer` · Category: `correctness`

Evidence: The new cleanup logic assigns values["pty_background_generation"] = None only when new_status is in {"pending", "failed"}. The same stale-state path can leave the marker attached when a task is reconciled as completed or cancelled.

Impact: A terminal task can continue to project background PTY activity to API/frontend consumers, causing later messages to be routed toward a dead session or incorrectly blocked even though its runtime has been released.

Required fix: Clear pty_background_generation whenever stale-runtime reconciliation reaches any terminal, ownerless task state, including completed and cancelled; preserve it only while an exact live runtime owner is proven.

Verification: Add cleanup tests for completed and cancelled tasks with a retained PTY marker and assert the marker is null and subsequent queued-message admission is unblocked.

Finding fingerprint: `9bc7eaa11a08fb608b600c4e695cce268f5c8e3278aaa026df7f416810819688`

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     # 分布式 Worker（EC2 provisioning）；WORKER_ENABLED=false 时不 import
     "boto3>=1.34.0",
     # PTY 模式（持久交互会话）；缺失时优雅降级为 `claude -p`
-    "claude-pty @ git+https://github.com/zjw49246/Claude-Code-PTY",
+    "claude-pty @ git+https://github.com/Citrus086/Claude-Code-PTY@32f1e5ac5cb3d2f0eeecb1e6a7c8a5396bc6301f",
```
youchengsong, line 24: **[MEDIUM] Production PTY dependency is switched to an unreviewed fork**

Reviewer: `principal_engineer` · Category: `security`

Evidence: The patch changes the runtime dependency from github.com/zjw49246/Claude-Code-PTY to github.com/Citrus086/Claude-Code-PTY at a new commit, with the lockfile updated to the fork as well. The patch supplies no provenance, review, or compatibility verification for code executed in every deployment.

Impact: All installations now execute third-party fork code, creating supply-chain and availability risk and making the fix dependent on a repository outside the original dependency ownership.

Required fix: Use a reviewed upstream commit or formally vendored/reviewed fork with documented provenance and compatibility checks; keep pyproject.toml and uv.lock aligned to that canonical source.

Verification: In CI, install from a clean environment and verify the exact approved repository/commit, then run PTY launch/resume integration tests against that artifact.

Finding fingerprint: `41bbccdc8a9140932712fed32c47106f0c69f9a12d4392ce5bfacd28129efb09`

backend/services/dispatcher.py
```diff
@@ -22961,6 +23041,89 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
...
+                        None,
+                    )
+                    if lifecycle_generation is None or (
+                        lifecycle_generation.task_id != task.id
+                        or any(
+                            getattr(lifecycle_generation, field)
+                            != getattr(task, field)
+                            for field in (
+                                "worker_id",
+                                "shared_from_id",
```
youchengsong, line 23072: **[MEDIUM] Terminal timestamp prevents stale lifecycle reconciliation**

Evidence: The lifecycle generation is captured when the lifecycle is registered at dispatch, before terminal completion, but reconciliation requires every listed field—including `completed_at`—to equal the current terminal Task. A normal terminal transition changes `completed_at` from the launch-time value, causing this branch to return `True` instead of cancelling the stale lifecycle. The added test masks this by constructing an already-failed Task and only then calling `_task_lifecycle_generation(task)`, so both snapshots retain the same value.

Impact: After a lifecycle becomes stuck during terminal cleanup, queued follow-up messages can continue to be considered busy forever, defeating the recovery behavior this patch claims to add.

Required fix: Do not use terminal-mutated fields such as `completed_at` when matching the immutable launch generation; match only stable ownership/generation fields or preserve and compare an explicit immutable generation token.

Verification: Capture `_task_lifecycle_generation` while a Task is executing with `completed_at=None`, then transition it to a terminal status with a populated `completed_at` and an ownerless instance. Verify `_queued_task_has_live_generation` cancels and removes the stale lifecycle and returns `False`.

Finding fingerprint: `f67c206dfc96ca9992a7e3b188b6acc32f4ba6368c4311b43bea14df52e6b9a5`

youchengsong (commented):

## Reviewer summaries

Terminal PTY guard reconciliation still fails when a task retains a now-missing instance reference, allowing queued turns to remain blocked indefinitely.

The new 413 marker is accepted through a provider-agnostic context classifier, allowing Claude-specific rendered error text to trigger context recovery for other providers.

The stale-lifecycle recovery compares a launch snapshot with mutable terminal timestamps, while its regression test creates the snapshot only after terminalization. This can leave the production deadlock it intends to fix.

Reviewed commit: `6ee9c87bdec3fd4d8aa0dc1f6fe378afe7b7a4c1`.

backend/services/context_compaction.py
```diff
@@ -417,6 +477,8 @@ def is_context_window_exceeded(provider: str | None, *details: Any) -> bool:
     """Recognize provider text and Codex app-server's structured error code."""
 
     del provider  # Markers are intentionally valid for both supported CLIs.
+    if any(is_claude_request_too_large_413(detail) for detail in details):
+        return True
```
youchengsong: **[MEDIUM] Claude-specific 413 text is classified as context overflow for every provider**

Evidence: `is_context_window_exceeded` explicitly discards `provider` (`del provider`) and now returns `True` whenever any detail matches `is_claude_request_too_large_413`. That helper accepts the rendered string `Request too large (max ...MB). Double press esc ...` without any provider or Claude-envelope validation.

Impact: A Codex (or future provider) error/output containing this text is treated as a recoverable context-window failure, causing compaction and fresh-session retry when the failure may be unrelated. This can discard the active session state and change retry behavior for non-Claude tasks.

Required fix: Keep the 413 detection in Claude-specific call paths, or gate the rendered-text branch on `provider == "claude"` while preserving only genuinely provider-independent markers in `is_context_window_exceeded`.

Verification: Add a regression asserting `is_context_window_exceeded("codex", "Request too large (max 32MB). Double press esc to go back and try with a smaller file.")` is false, while the same input for Claude remains true.

Finding fingerprint: `10803272473702e1b7facd81568f95b628a7c9b2031d0b5dc5cb935c42f0b7b3`

backend/services/dispatcher.py
```diff
@@ -22917,8 +22946,44 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
...
+            await db.get(Instance, instance_id, populate_existing=True)
+            if instance_id is not None
+            else None
+        )
+        manager_running = bool(
+            instance_id is not None
+            and self.instance_manager.is_running(instance_id)
+        )
         session_id = task.session_id
         if session_id:
```
youchengsong: **[HIGH] Orphaned PTY guards are not reconciled when the retained instance row is missing**

Evidence: `terminal_ownerless` is true only when `instance_id is None` or when `instance is not None` and the instance has an ownerless terminal snapshot. A task with a non-null `task.instance_id` whose instance lookup returns `None` therefore never enters the `reconcile_orphaned_pty_runtime_guard` path.

Impact: A stale task/session guard can remain indexed after its Instance row is deleted or otherwise unavailable. `_queued_task_has_live_generation` continues treating the session guard as live, so later queued user messages can wait indefinitely despite no durable or runtime owner existing.

Required fix: Treat a missing Instance row as ownerless when the task is terminal (or explicitly reconcile the dangling instance reference) before deciding whether to retain the PTY guard. Preserve the existing fail-closed behavior when any runtime evidence is present.

Verification: Add a dispatcher regression test with a terminal task retaining a nonexistent `instance_id`, a stale PTY handoff/background marker, and no live manager runtime; assert reconciliation is invoked and `_queued_task_has_live_generation` returns false.

Finding fingerprint: `7f46db45bf76c2f5cf2cab9e24e3f11482dd5c23b65b9bf07b4045af74984078`

backend/services/dispatcher.py
```diff
@@ -22961,6 +23041,89 @@ async def _queued_task_has_live_generation(self, db, task_id: int) -> bool:
...
+                    and instance.current_plan_run_id is None
+                    and instance.pid is None
+                    and not manager_running
+                ):
+                    lifecycle_generation = getattr(
+                        lifecycle,
+                        "_ccm_task_generation",
+                        None,
+                    )
+                    if lifecycle_generation is None or (
```
youchengsong: **[MEDIUM] Terminal timestamp comparison prevents stale lifecycle recovery**

Evidence: The lifecycle generation is captured when the lifecycle is registered, before it runs, but reconciliation later requires `completed_at` (and other listed fields) to equal the now-terminal Task. Normal terminalization can change `completed_at`, causing this branch to return `True` rather than cancel the dead lifecycle. The added test does not exercise that transition: it creates an already-`failed` Task and then derives `lifecycle_generation` from that same terminal row, so all compared fields necessarily match.

Impact: A lifecycle suspended during terminal cleanup can continue to be reported as live forever, blocking subsequent queued chat messages—the exact production failure this recovery path claims to address.

Required fix: Do not use lifecycle-mutated terminal fields such as `completed_at` when identifying the immutable generation. Compare only stable generation/ownership fields, or retain and validate an explicit immutable generation token that remains valid across terminalization.

Verification: Capture the lifecycle generation from an executing Task with `completed_at=None`, then terminalize the persisted Task and set a non-null `completed_at` while leaving its lifecycle suspended. Assert `_queued_task_has_live_generation` cancels and settles that lifecycle, removes it from `_running_tasks`, and returns `False`; also retain mismatch cases for retry/turn generation.

Finding fingerprint: `2172f043f60a3302451e1e3e20c5bce5d59005e4112841bba10b6bec661aa073`

